PR #3, opened 2015-01-31
Use environment variables to set user permissions.

bgruening:
Hi @jmchilton,

I found this nice trick to use environment variables inside of `proftpd.conf`.
I will need this for galaxy-stable. If you like it we can also replace `proftpd_files_dir`.


jmchilton:
Awesome - thanks for tracking down the problem.

PR #4, opened 2015-02-01
Expose the Galaxy report webapp on port 9003 via nginx.

bgruening:
* use nginx proxy
* serve static content

### DIFF
--- a/tasks/galaxy_root.yml
+++ b/tasks/galaxy_root.yml
@@ -9,7 +9,7 @@
   sudo: True
   sudo_user: "{{ galaxy_user_name }}"
 
-- name: "Install watchdog for galaxy"
+- name: "Install uwsgi for galaxy"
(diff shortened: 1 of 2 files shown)


jmchilton:
I wonder if this could be improved on:
 - if you are deploying a personal/lab serer, then it would be better if it were just if `:80/reports` was proxied to the reports serer instead of `:9003/` - so that you wouldn't need to open a second port.
 - if you have a public server - I am not sure this current version is very helpful since you can only expose it or not - it would be better to have it on `:80/reports` with some sort of password protection.

So I think the right default configuration might be `:80/reports` that is password protected if some flag is set (maybe an environment variable or maybe the presence of a credentials file).

bgruening:
Yes, this sounds good, did not know you can password protect reports.
I think everything is working now. This is a bunch of really cool changes John, thanks a lot.
Let's improve things even further.

jmchilton:
Awesome thanks so much for fixing up the things I broke.

PR #5, opened 2015-03-05
Add new nginx location so we can set welcome_url to /etc/galaxy/

bgruening:
This is needed for https://github.com/bgruening/docker-galaxy-stable/issues/31 and potential debian packages.


jmchilton:
Awesome, thanks!

PR #7, opened 2015-03-06
nginx_proxy_reports_port is not needed anymore. 

bgruening:
We can simply proxy to `galaxy_reports_port`. This will change the default port of reports from `9003` back to `9001`.


jmchilton:
Awesome, thanks.

PR #9, opened 2015-03-21
Welcome page fix

bgruening:
This should fix https://github.com/bgruening/docker-galaxy-stable/issues/48

### DIFF
--- a/templates/nginx.conf.j2
+++ b/templates/nginx.conf.j2
@@ -114,9 +114,9 @@ http {
             alias /;
         }
 
-        # this is needed if 'welcome_url' is set to /etc/galaxy
-        location /etc/galaxy {
-            alias /etc/galaxy;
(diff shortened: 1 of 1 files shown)


jmchilton:
I will admit to not understanding the fix really - but is that exposing that whole directory? You probably don't want to do that right since it might have for instance a ``master_api_key`` in the galaxy.ini file. 

If this solution works - it seems like you could create an ``/etc/galaxy/web`` maybe and change the value in the init file and alias to ``/etc/galaxy/web``.

bgruening:
I was not able to get `/etc/galaxy` listed. But better save than sorry. Thanks for the hint.

bgruening:
@jmchilton do you think this is ok now?

jmchilton:
Sure - thanks for making that change. Does the initial startup script need to be modified also?

PR #12, opened 2015-06-24
[WIP] Standing PR containing enhancements for GCC 2015

jmchilton:
- Bring in runtime scripts from docker-galaxy-stable.
 - Infrastructure for checking a directory and changing the UID based on that. Needed for https://github.com/galaxyproject/planemo/issues/212.
 - Infrastructure for managing a tool shed in VM/container.
 - Other small tweaks to supervisor configuration (galaxy needs more retries than it use to because postgres is taking awhile to setup - not sure what that is about or what changed).

### DIFF
--- a/defaults/main.yml
+++ b/defaults/main.yml
@@ -56,6 +56,9 @@ galaxy_paste_handlers: false
 galaxy_web_processes: 2
 galaxy_handler_processes: 2
 
+startup_export_user_files: true
+startup_chown_on_directory: false
+
 # Nginx configuration.
 nginx_conf_path: /etc/nginx/nginx.conf
 nginx_conf_directory: /etc/nginx/conf.d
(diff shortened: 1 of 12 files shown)


bgruening:
@jmchilton I like this. At some point galaxy-stable is nearly empty :)

jmchilton:
@bgruening Ah yes - there is that. Essentially my career has devolved into coming up with different ways of taking credit for things you do - and I am not stopping until someone complains.

jmchilton:
Okay - I just had a successful docker run that updated the files, had a toolshed with the proper categories, etc.... and it looks like the virtualbox run is going to succeed as well. I am going to merge this down now. docker-galaxy-stable will be fine because it is pinned to an older version but at some point (GCC?) we should find some time to migrate it to the newest version and remove the duplication.

bgruening:
@jmchilton I don't care about credits. It's just nice to see it is of any use. Thanks for working on this.

afgane:
> we should find some time to migrate it to the newest version and remove the duplication.

It'd be good to do this across all the roles to unify the approaches for building various infrastructures and consolidate as many parts as possible. This would not only help all of us reduce the amount of duplicate effort but also help people that build the components to have a more consistent experience. 

Unfortunately I won't be at GCC but would like to get together at some point to talk this over? I feel we should be able to have a number of reusable roles that can then be assembled in standalone playbooks where each playbook targets a specific need/infrastructure but the resultant Galaxy is largely the same.

bgruening:
@afgane yes please! You have a few nice scripts that would be nice to reuse!

PR #14, opened 2015-10-07
Adding ability to configure docker mounted volumes galaxy_docker_volumes variable

kellrott:


jmchilton:
Can you add a default for ``galaxy_docker_volumes`` in https://github.com/galaxyproject/planemo-machine/blob/master/group_vars/all.

kellrott:
Yes. I have that already, but was going to wait for this PR to go through before updating the module link and adding the default.

jmchilton:
https://github.com/galaxyproject/ansible-galaxy-extras/blob/master/defaults/main.yml

PR #15, opened 2015-10-20
Extra nginx config

kellrott:
Adding ability to define additional nginx config files to be included in the server definition.


jmchilton:
Awesome sauce, looks good to me.

PR #18, opened 2015-11-19
Remove fetching eggs from slurm task

bgruening:
@natefoo @jmchilton is it save to remove this portion. It keeps failing on Docker:dev.
I assume that wheels will take care of this?


jmchilton:
Seems like this should be contigent on targeting Galaxy dev, at least through the beginning of next year.

bgruening:
This fix from @natefoo is much nicer: https://github.com/galaxyproject/ansible-galaxy-extras/pull/16/files
Thanks for merging.

PR #20, opened 2015-11-19
use venv in startup.sh

bgruening:


jmchilton:
Thanks @bgruening, sorry for the lag.

PR #21, opened 2015-11-20
Remove obsolete --plugin python from supervisor template, and enforce…

fabiorjvieira:
… stopping of services before removal.

We are using this role on a VM and in the cloud.
This way we do not need to stop the services manually before launching them in supervisor.


jmchilton:
@fabiorjvieira Sorry I missed this completely, please ping me if you open additional PRs and it takes more than a week to respond. I believe @mvdbeek's changes include all of these enhancements.

PR #22, opened 2015-11-20
Update to use the latest psutil version

bgruening:


jmchilton:
We keep changing psutil, can we just make this code independent of the version we use - with a try except sort of thing?

jmchilton:
Specifically, the file use to look like this right - 
https://github.com/galaxyproject/ansible-galaxy-extras/commit/3736c1d36f9dfa4d832f5b5fc22fbc142b090ea0

bgruening:
@jmchilton good catch! I have fixed it. Thanks!

jmchilton:
Awesome sauce, thanks!

PR #27, opened 2015-11-29
allow dynamic number of Galaxy handlers

bgruening:
@jmchilton what do you think? This is the most elegant solution I can think of. supervisor does not support `ENV_` in numprocs.

### DIFF
--- a/templates/startup.sh.j2
+++ b/templates/startup.sh.j2
@@ -4,6 +4,9 @@ cd {{ galaxy_root }}
 . {{ galaxy_root }}/.venv/bin/activate
 umount /var/lib/docker
 
+# Set number of Galaxy handlers via GALAXY_HANDLER_NUMPROCS or default to 2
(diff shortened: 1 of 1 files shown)


jmchilton:
Is there something missing here - where ini_file coming from? The existing playbook stuff... that feels kind of gross right, we currently don't have any runtime dependency on ansible.

jmchilton:
Can this just be replaced with a sed statement?

bgruening:
@jmchilton `ini_file` is a ansible module to modify ini files: http://docs.ansible.com/ansible/ini_file_module.html

I thought this is way more elegant than sed or any python script we need to write. `sed` will be difficult and probably ugly. Ini file modifications are not so straight forward imho. Thats probably also the reason ansible has a module for it.

This with the runtime dependency is a bummer. But I guess since we use the galaxy-tools role / playbook we have now a run time dependency on it, right?

https://github.com/bgruening/docker-galaxy-stable/blob/dev/galaxy/install_tools_wrapper.sh#L43

Nevertheless, do you think it's worth to have a ansible dependency? Otherwise we need to write a python script that does exactly the same as ansible ini_file.

jmchilton:
This is fine then, I didn't understand what was happening - that looks perfect.

PR #30, opened 2015-12-25
Adjust for available slurm packages in wily, make munge key creation …

mvdbeek:
…idempotent, and stop and remove postgres if it is supposed to be controlled by supervisor.


jmchilton:
Fantastic, thanks a bunch @mvdbeek!

bgruening:
Nice one!

PR #31, opened 2015-12-26
plugin python is built into core uwsgi if installed via pip.

mvdbeek:
[Distros should package uWSGI in a modular way, with each feature as a plugin. But when you install using language specific ways (pip, gem...) the relevant language is embedded, so you do not need to load the plugin] (http://stackoverflow.com/a/15941424)

I tested this with the stable branch of @bgruening docker-galaxy-stable, where this works fine.
(uwsgi is installed into the virtualenv environment using pip)
If we do have `--plugin python` in all of our current project, starting galaxy_web fails with
```
open("./python_plugin.so"): No such file or directory [core/utils.c line 3675]
!!! UNABLE to load uWSGI plugin: ./python_plugin.so: cannot open shared object file: No such file or directory !!!
``` 
This has also occured [here](https://github.com/bgruening/docker-galaxy-stable/issues/76)
It also supersedes PR #21

### DIFF
--- a/templates/supervisor.conf.j2
+++ b/templates/supervisor.conf.j2
@@ -68,7 +68,7 @@ priority        = 200
 
 [program:galaxy_web]
 {% if galaxy_uwsgi %}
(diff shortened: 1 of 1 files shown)


mvdbeek:
@jmchilton thanks!

jmchilton:
@mvdbeek No, thank you this is awesome. I never even noticed #21 :cold_sweat:, too many projects I seem to miss PRs on things other than Galaxy core.

jmchilton:
@mvdbeek I have added you to the Galaxy "docker-devops" team that has write access to a few projects including this one. Feel free to commit directly or continue to open PRs of course, whichever you'd prefer.

mvdbeek:
@jmchilton no worries, i also miss notifications on github, and I'm involved in far less than what you must be seeing.
Thanks a lot for all the things you're doing!

PR #32, opened 2015-12-26
Make specifying uwsgi parameters with environmental variables optional

mvdbeek:
using `galaxy_uwsgi_static_conf`.
Use `postgresql_version` (also used in https://github.com/galaxyproject/ansible-postgresql) to determine location of postmaster binary and `supervisor_postgres_database_path`.


mvdbeek:
@bgruening is this ok for you? 
I understand the utility of `-e UWSGI_PROCESSES=N` when creating the docker container, but for maintaining a running machine with ansible this is a bit cumbersome.
Using environmental variables remains the default, and you can switch to "static" using `galaxy_uwsgi_static_conf`.

bgruening:
@mvdbeek  this looks great to me!
:+1:

mvdbeek:
@bgruening thanks!

PR #33, opened 2016-01-02
Fix for expanduser - conda

mvdbeek:
The new conda dependency resolver will expand '~', which points to root (the user that starts supervisor).
Circumvent the problem by specifying HOME in the environmental variables.

### DIFF
--- a/defaults/main.yml
+++ b/defaults/main.yml
@@ -53,6 +53,9 @@ galaxy_database_connection: "postgres://{{ galaxy_user_name }}@localhost:{{ gala
 galaxy_uwsgi: true
 uwsgi_port: 4001
 
+# Set to true to write thread/process number into supervisor config. If false, uses environmental variables.
+galaxy_uwsgi_static_conf: false
+
 # Set the following to true - to run paste processes for Galaxy
 # handlers (should not longer be needed).
 galaxy_paste_handlers: false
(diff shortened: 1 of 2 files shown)


mvdbeek:
@jmchilton it's not unlikely other users will encounter this problem. Should galaxy perhaps check if ~ is writeable (https://github.com/galaxyproject/galaxy/blob/dev/lib/galaxy/tools/deps/conda_util.py#L100)?

bgruening:
I think conda should only write to `tool_dependency_dir` and not to some home directory. I'm not sure what happens if the compute node does not have $HOME set. Can we use `tool_dependency_dir` instead of `~`?

mvdbeek:
Yep, I guess that would be a better (upstream?) solution.

bgruening:
Yes, let see what John thinks about this. I kind of like that we do not set the HOME here and catch such errors. Compute nodes don't have a $HOME either and we should not rely on it.

jmchilton:
I don't know how to overwrite conda's rc location, it seems if these things can be configured via environment variable it is not mentioned anywhere. I'll open a PR against Galaxy right now to delay writing that file until an actual install is going to occur and I'll check if ~ is writieable. Longer term though, it would be nice if we could configure conda more... precisely.

jmchilton:
Looks like we can override the location with CONDARC.

https://github.com/conda/conda/blob/master/conda/config.py#L102

bgruening:
@jmchilton perfect we could default to `tool_dependency_dir`/`__conda__`

jmchilton:
Should be fixed once @bgruening merges https://github.com/galaxyproject/galaxy/pull/1409. Sorry about that.

mvdbeek:
Thanks @jmchilton and @bgruening , so I'll close this!

PR #34, opened 2016-01-02
Set galaxy_uwsgi_static_conf to false by default.

mvdbeek:
set PYTHONHOME to galaxy_venv_dir.

### DIFF
--- a/defaults/main.yml
+++ b/defaults/main.yml
@@ -53,6 +53,9 @@ galaxy_database_connection: "postgres://{{ galaxy_user_name }}@localhost:{{ gala
 galaxy_uwsgi: true
 uwsgi_port: 4001
 
+# Set to true to write thread/process number into supervisor config. If false, uses environmental variables.
+galaxy_uwsgi_static_conf: false
+
 # Set the following to true - to run paste processes for Galaxy
 # handlers (should not longer be needed).
 galaxy_paste_handlers: false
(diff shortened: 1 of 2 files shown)


bgruening:
This one is a must `galaxy_uwsgi_static_conf: false` and an important bugfix.
I guess the `PYTHONHOME` does not hurt, I'm just wondering why this is needed.

mvdbeek:
I'm not quite sure either why PYTHONHOME is needed, but when running eggless galaxy all imports are failing. Normally this is supposed to be taken care of when you call the python within the virtualenv ... .

bgruening:
With eggless you mean Galaxy on wheels :)?
Can we try to fix this upstream? @natefoo can probably enlighten us here.
Do you mind to create a separate PR for the bugfix?

mvdbeek:
Sure, no problem.

mvdbeek:
closing this in favor of PR #35 and raising an issue for the PYTHONHOME problem.

natefoo:
We use `PYTHONHOME` (actually, `--virtualenv`, which is an alias for the option version of that var) for the tool shed as well since uWSGI is installed via apt on that system. If uwsgi is installed in Galaxy's virtualenv, the option is not necessary.

[Related change](https://github.com/galaxyproject/usegalaxy-playbook/commit/b4841230c39de5e87136d371a9a7366b8055689d)

mvdbeek:
thanks @natefoo, this is still weird then. Here we're already using `{{ galaxy_venv_dir }}/bin/uwsgi`, yet somehow we seem to need to set `PYTHONHOME`. @bgruening should we also use `--virtualenv` ?

mvdbeek:
This is superceded by PR #35 and PR #36

PR #36, opened 2016-01-05
venv fixes for galaxy-on-wheels and improvements for running outside of docker

mvdbeek:
- 2d7774e stop supervisor during service management.

- 851f58c since the ENV variables are not available during `docker build` we don't start supervisor back up if we use ENV vars to control supervisor managed programs. [docker-galaxy-stable continues to work]

- 776ad83 on real machines (VM or bare metal), we need to create /var/run/postgresql with the proper permissions once after boot. This will run the install command once (`startsecs = 0`) after starting supervisor if we are not in docker. Same thing happens in init.d script supplied with ubuntu.

-  3b3a964 we still need PYTHONHOME for the handler, else handler jobs will fail while looking for sqlalchemy.orm.

### DIFF
--- a/tasks/supervisor.yml
+++ b/tasks/supervisor.yml
@@ -10,6 +10,9 @@
 - name: Create Galaxy configuration file
   template: src=supervisor.conf.j2 dest={{ supervisor_conf_path }}
 
+- name: Stop supervisor
+  service: name=supervisor state=stopped
+
 - name: Stop services before removing them.
   service: name={{ item }} state=stopped enabled=no
   with_items:
@@ -36,3 +39,6 @@
   with_items:
   - postgresql
   when: supervisor_manage_postgres
+
+- name: Start supervisor
+  service: name=supervisor state=started
--- a/tasks/supervisor.yml
+++ b/tasks/supervisor.yml
@@ -40,5 +40,7 @@
   - postgresql
   when: supervisor_manage_postgres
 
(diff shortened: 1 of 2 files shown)


mvdbeek:
@bgruening test using docker-galaxy-stable is here:
https://travis-ci.org/mvdbeek/docker-galaxy-stable/builds/100308140

mvdbeek:
Unfortunately the `pattern="no_pattern"` is not working as documented ... it does take precedence over status, and therefore the service is never properly stopped :(. A better workaround is to completely take down supervisor before stopping the services and starting after the service management is done.

mvdbeek:
Alright, latest test results are [here] (https://travis-ci.org/mvdbeek/docker-galaxy-stable/builds/100352577), PR description is updated too. This should be ready from my side, @bgruening any comments?

bgruening:
This postgresql change could touch things from @jmchilton or @natefoo. So I will leave them to merge this one.

Thanks @mvdbeek for using these pieces.

bgruening:
Thanks @jmchilton!

jmchilton:
Seems solid, I haven't rebuilt a planemo machine but next time I'll do so and if there are problems they will likely be small - more likely though this fixed planemo machine for 16.01 - so thanks :smile:.

mvdbeek:
thanks @bgruening @jmchilton !

PR #37, opened 2016-01-08
Make start configurable and allow proftpd to be NAT'ted

mvdbeek:
If proftpd is running in a NAT setting, `MasqueradeAddress` needs to be set to the public IP clients use to connect in order to have passive mode working. This is configurable and needed for example in amazon's default VPC.
In addition the passive ports are configurable as well as which services start by default.


mvdbeek:
building docker-galaxy-stable [here] (https://travis-ci.org/mvdbeek/docker-galaxy-stable/builds/101077366)

mvdbeek:
unfortunately dig requires an open udp/53, which we don't seem have on the french academic cloud.
I could make this configureable, with a default to something like `curl ipv4.icanhazip.com` ... less than ideal, but I don't know another way to get the public ip address.

bgruening:
I probably don't get the problem here, but does your public IP address change?
I like the other changes!

mvdbeek:
Yes, if you publish your AMI and boot it up again it will be attributed a different public ip.

mvdbeek:
@bgruening 
This approach is working now, and using `public_ip` you can set it on the playbook level to an IP or a command that returns an IP.

defaults/main.yml
```diff
@@ -99,6 +99,12 @@ proftpd_db_username: galaxy
 proftpd_db_password: galaxy
 proftpd_welcome: "Public Galaxy FTP"
 proftpd_files_dir: /export/galaxy-central/database/ftp
+# Set masquearade to true if host is NAT'ed.
+proftpd_nat_masquerade: false
+proftpd_passive_port_low: 30000
+proftpd_passive_port_high: 40000
+# public_ip refefers to the ip that clients use to establish an ftp connection
+public_ip: "`curl ipv4.icanhazip.com`"
```
bgruening: Can we add to the documentation that this is only relevant if `proftpd_nat_masquerade` is true?

bgruening:
Like it! @natefoo could you please look over it and merge if you feel comfortable?

defaults/main.yml
mvdbeek: Will do.

defaults/main.yml
```diff
@@ -120,6 +126,8 @@ supervisor_galaxy_startsecs: 20
 supervisor_galaxy_startretries: 15
 
 supervisor_ie_proxy_autostart: false
+supervisor_proftp_autostart: false
```
natefoo: Could this be made `supervisor_proftpd_autostart`?

defaults/main.yml
natefoo: On EC2 this can be obtained from the [ec2_facts module](http://docs.ansible.com/ansible/ec2_facts_module.html) or `ec2metadata --public-ipv4` if using the public images.

defaults/main.yml
mvdbeek: sure

mvdbeek:
sorry, i got the merge wrong .... argh. need to clean this up first, but thanks for looking over it @bgruening  @natefoo

defaults/main.yml
mvdbeek: Ah, that's much nicer ... i will go with `ec2metadata --public-ipv4` as a default, since we may need to run this upon boot.

mvdbeek:
@bgruening @natefoo from my side this is okay, if you prefer I can squash the commits.

natefoo:
This is great, thanks! I also appreciate that you converted some usage of shell to proper Ansible modules along the way.

jmchilton:
"proper" :unamused:.. I see you @natefoo.

natefoo:
@jmchilton Almost called you out by name but decided to go the more underhanded route. =D

PR #39, opened 2016-01-11
Detect if proftp, posgresql or nginx was set to be controled by supervisor

fabiorjvieira:
If proftp, posgresql, slurm or nginx are not controlled by supervisor, the previous tasks/supervisor.yml with remove them from the auto-start status and possibly crash other services that depend on them.

### DIFF
--- a/tasks/supervisor.yml
+++ b/tasks/supervisor.yml
@@ -10,11 +10,20 @@
 - name: Create Galaxy configuration file
   template: src=supervisor.conf.j2 dest={{ supervisor_conf_path }}
 
+- name: Stop services before removing them.
+  service: name={{ item }} state=stopped
+  with_items:
+  - nginx
+  - uwsgi
+  - proftpd
+  - munge
+  - slurm-llnl
+
 - name: Disable init scripts for stuff managed by supervisor.
   command: update-rc.d -f {{ item }} remove || true
   with_items:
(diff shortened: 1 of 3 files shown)


templates/supervisor.conf.j2
```diff
@@ -56,7 +56,7 @@ priority        = 100
 {% if supervisor_manage_proftp %}
 [program:proftpd]
 command         = /usr/sbin/proftpd -n -c {{proftpd_conf_path}}
-autostart       = false
+autostart       = true
```
mvdbeek: @fabiorjvieira can you make this configurable with default no?
sth. like {{ supervisor_docker_autostart }}

tasks/supervisor.yml
```diff
@@ -16,30 +16,40 @@
...
 
-- name: Stop services before removing them.
+- name: Stop postgresql before removing it from services.
   service: name={{ item }} state=stopped enabled=no
   with_items:
-  - postgresql
+    - postgresql
   when: supervisor_manage_postgres
 
+- name: Stop proftpd service before removing them.
```
mvdbeek: @fabiorjvieira while you're at it, "Stop and remove postgresql" is more strightforward, right?

mvdbeek:
To me this is good, but could you squash the commits?

PR #41, opened 2016-01-11
Detect if proftp, posgresql, slurm or nginx were set to be controled by supervisor

fabiorjvieira:
Changes:
   Detect if proftp, posgresql, slurm or nginx were set to be controled by supervisor and deal with the systems services accordingly. 
  The autostart of proftpd is now an option.

### DIFF
--- a/defaults/main.yml
+++ b/defaults/main.yml
@@ -129,7 +129,6 @@ supervisor_galaxy_startsecs: 20
 supervisor_galaxy_startretries: 15
 
 supervisor_ie_proxy_autostart: false
-supervisor_proftpd_autostart: false
 supervisor_reports_autostart: false
 
 supervisor_docker_autostart: false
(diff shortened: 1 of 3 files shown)


defaults/main.yml
```diff
@@ -138,3 +139,5 @@ supervisor_webserver: true
 supervisor_webserver_port: "0.0.0.0:9002"
 supervisor_webserver_username: null
 supervisor_webserver_password: changeme
+
+supervisor_proftpd_autostart: false
```
mvdbeek: @fabiorjvieira this is a duplicate with https://github.com/fabiorjvieira/ansible-galaxy-extras/blob/pr_services/defaults/main.yml#L132

fabiorjvieira: Thanks @mvdbeek, fixed. I also must squash some commits.

Issue #17, opened 2015-11-16
additional support for pbs-clusters


matthdsm:
Hi, 

I made some changes to this role to include support for pbs-torque clusters, would you like me to send a pull request so these changes can be checked out/imported into the project?

Greetings
Matthias

mvdbeek:
@matthdsm it's a bit late as a response, but personally I think that would be a good idea!

bgruening:
For the Docker container it would be nice to switch it at run-time as well.

matthdsm:
The pull request got merged, so I'll close this issue. 

Matthias

PR #45, opened 2016-01-15
Added support for pbs-torque clusters

matthdsm:
pull request created in follow up of #17 
I've used this edited role to deploy our local galaxy instance and I'm not having any issues.

It's possible however there needs to be some more work done to make it applicable for a wider audience.

Greetings
Matthias

### DIFF
--- a/defaults/main.yml
+++ b/defaults/main.yml
@@ -6,7 +6,6 @@ galaxy_extras_config_nginx_upload: true
 galaxy_extras_config_proftpd: true
 galaxy_extras_config_slurm: true
 galaxy_extras_config_pbs: false
-pbs_server_name: pbsqueue
 galaxy_extras_config_supervisor: true
 galaxy_extras_config_galaxy_root: true
 galaxy_extras_config_galaxy_job_metrics: true
@@ -14,6 +13,9 @@ galaxy_extras_config_uwsgi: true
 galaxy_extras_config_ie_proxy: true
 galaxy_extras_config_scripts: true
 galaxy_extras_config_startup: true
+
+# set the FQDN for the pbs server, only used when galaxy_extras_config_pbs: true
+pbs_server_name: pbsqueue
 
 # Install OS packages corresponding to each service being configured.
 galaxy_extras_install_packages: false
--- a/defaults/main.yml
+++ b/defaults/main.yml
@@ -2,6 +2,7 @@
 # https://github.com/bgruening/docker-galaxy-stable.
 
 galaxy_extras_config_nginx: true
+galaxy_extras_config_nginx_upload: true
 galaxy_extras_config_proftpd: true
 galaxy_extras_config_slurm: true
 galaxy_extras_config_pbs: false
--- a/defaults/main.yml
+++ b/defaults/main.yml
@@ -4,6 +4,8 @@
 galaxy_extras_config_nginx: true
 galaxy_extras_config_proftpd: true
 galaxy_extras_config_slurm: true
+galaxy_extras_config_pbs: false
+pbs_server_name: pbsqueue
 galaxy_extras_config_supervisor: true
 galaxy_extras_config_galaxy_root: true
 galaxy_extras_config_galaxy_job_metrics: true
--- a/defaults/main.yml
+++ b/defaults/main.yml
@@ -106,6 +106,7 @@ supervisor_manage_nginx: true
 supervisor_manage_reports: true
 supervisor_manage_ie_proxy: true
 supervisor_manage_docker: true
+supervisor_manage_nodeproxy: true
 supervisor_manage_toolshed: false
 
 supervisor_galaxy_startsecs: 20
--- a/defaults/main.yml
+++ b/defaults/main.yml
@@ -25,8 +25,10 @@ galaxy_extras_apt_package_state: present
 
 galaxy_extras_ie_fetch_ipython: false
 galaxy_extras_ie_ipython_image: bgruening/docker-ipython-notebook:dev
+galaxy_extras_ie_fetch_jupyter: false
+galaxy_extras_ie_jupyter_image: bgruening/docker-jupyter-notebook:16.01
 galaxy_extras_ie_fetch_rstudio: false
-galaxy_extras_ie_rstudio_image: erasche/docker-rstudio-notebook:dev
+galaxy_extras_ie_rstudio_image: erasche/docker-rstudio-notebook:15.10
 
 galaxy_docker_enabled: false
 galaxy_docker_sudo: false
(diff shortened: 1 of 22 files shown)


tasks/pbs.yml
```diff
@@ -0,0 +1,22 @@
...
+  when: galaxy_extras_install_packages
+
+- name: "Fetch DRMAA egg for Galaxy"
+  shell: "{{ galaxy_venv_dir }}/bin/python {{ galaxy_root }}/scripts/fetch_eggs.py -e drmaa -c {{ galaxy_config_file }}"
+
+- name: "Install Galaxy job conf"
+  template: src=job_conf.xml.j2 dest={{galaxy_job_conf_path}}
+
+- name: "Set PBS/torque server name"
+  shell: "echo {{ pbs_server_name }} >> /etc/torque/server_name"
```
mvdbeek: this could be [lineinfile](https://docs.ansible.com/ansible/lineinfile_module.html), right?

mvdbeek:
@matthdsm Great, this looks really nice, thank you! The only thing I'm worried about is `become` ... I'm not sure we want break compatibility with ansible-1.X already, given that 2.0 is out just a few days ago and it does support the old `sudo`/`sudo_user` combination. Also you'll have to merge with the current master.
Any other thoughts @bgruening @natefoo ?

defaults/main.yml
```diff
@@ -2,8 +2,11 @@
 # https://github.com/bgruening/docker-galaxy-stable.
 
 galaxy_extras_config_nginx: true
+galaxy_extras_config_nginx_upload: true
 galaxy_extras_config_proftpd: true
 galaxy_extras_config_slurm: true
+galaxy_extras_config_pbs: false
+pbs_server_name: pbsqueue
```
mvdbeek: `pbs_server_name` corresponds to the name/fqdn/ip of PBS-cluster head node?
could you add a comment here?

tasks/pbs.yml
matthdsm: I'm not really sure. I think I used this formulation because the "server_name" file doesn't get created when you install the torque package. 
lineinfile can't create any files or can it?

mvdbeek: `state=present create=yes` should work according to the doc ...

matthdsm: you're right, I'll start editing the code when you're done going through it ;)

mvdbeek: I think I'm OK, but I can't guarantee you'll not have comments by other people ;)

matthdsm: I'll also wait a day, so our friends on the other side of the globe get a chance to review it.

tasks/pbs.yml
```diff
@@ -0,0 +1,22 @@
+- name: Install pip
```
bgruening: Why is this needed. You should be in a virutalenv and this should have pip installed, isn't it?

matthdsm: True! I'll remove it when I start editing

natefoo: @jmchilton wrote slurm.yml. It appears to be installed there so that we can subsequently `pip install psutil`, which isn't needed for Torque, so this can go.

tasks/pbs.yml
```diff
@@ -0,0 +1,22 @@
...
+
+- name: Install PBS/torque system packages
+  apt: pkg={{ item }} state={{ galaxy_extras_apt_package_state }}
+  with_items:
+    - torque-client
+    - pbs-drmaa-dev
+    - python-psutil
+  when: galaxy_extras_install_packages
+
+- name: "Fetch DRMAA egg for Galaxy"
```
natefoo: Galaxy is switching from eggs to wheels (16.01 will be the first release using wheels), see https://github.com/galaxyproject/ansible-galaxy-extras/blob/master/tasks/slurm.yml#L46 for how this is handled in the slurm tasks.

matthdsm: Hi nate, thanks for chiming in.
What will be the implications of the move to python wheels in 16.01? I don't think there will be any issues between slurm and torque, since in the current configuration torque also uses DRMAA.

M

natefoo: Right, you should only need to incorporate the recent changes from `slurm.yml` that handle the move to wheels.

matthdsm:
I think all issues should be resolved. If any more should pop up, let me know! Otherwise, merge away!

M

bgruening:
@matthdsm can you please rebase this PR. Thanks :)!!!

matthdsm:
Done! The tests seem to be running.

Matthias

Op 20 jan. 2016, om 15:52 heeft Björn Grüning <notifications@github.com<mailto:notifications@github.com>> het volgende geschreven:


@matthdsm<https://github.com/matthdsm> can you please rebase this PR. Thanks :)!!!

tasks/galaxy_root.yml
```diff
@@ -1,10 +1,16 @@
 # TODO: following task is kind of a crutch - shouldn't be needed.
 - name: "Run common_startup.sh to create welcome.html from sample"
   shell: "cd {{ galaxy_root}} && scripts/common_startup.sh"
-  sudo: True
-  sudo_user: "{{ galaxy_user_name }}"
+  become: yes
+  become_method: sudo
 
 - name: "Install watchdog for galaxy"
+<<<<<<< HEAD
```
bgruening: There is a merge conflict here ...

tasks/uwsgi.yml
```diff
@@ -1,4 +1,10 @@
 - name: "Install uwsgi for galaxy"
+<<<<<<< HEAD
+  shell: ". {{ galaxy_venv_dir }}/bin/activate && pip install uwsgi"
```
bgruening: Here as well ...

templates/nginx.conf.j2
```diff
@@ -129,11 +129,15 @@ http {
             alias {{ galaxy_root }}/config/plugins/visualizations/$vis_name/static/$static_file;
         }
 
+<<<<<<< HEAD
```
bgruening: And here ...

matthdsm:
Sorry, I seem to have been too hasty with pushing the code.
I resolved the merge conflicts.

Matthias

natefoo:
@matthdsm could you do the reversion from `become` and `become_user` to `sudo` and `sudo_user` as explained by @mvdbeek please?

templates/job_conf.xml.j2
```diff
@@ -1,36 +1,53 @@
...
+  {% endfor %}
 {% endif %}
     </handlers>
     <destinations default="cluster">
+{% if galaxy_extras_config_slurm == True %}
         <destination id="cluster" runner="slurm">
-{% if galaxy_source_shellrc %}
-            <env file="{{ galaxy_user_shellrc }}" />
+  {% if galaxy_source_shellrc %}
+          <env file="{{ galaxy_user_shellrc }}" />
```
mvdbeek: @matthdsm is this duplication necessary?
Is this to be able to have both slurm and pbs configured at the same time?

matthdsm: Good eye, none of the duplicated code was really needed. I removed it and added a conditional instead.

mvdbeek:
Okay, thanks a lot @matthdsm . If noone objects I will merge this tomorrow.

matthdsm:
Huraah, I’ll be my first contribution to something open source!
I’m a big boy now! :)

mvdbeek:
This calls for a :beer:   :laughing: !

martenson:
Thank you very much @matthdsm !

bgruening:
Oh this is awesome! Looking forward to offer this in the Docker Image :)
Thanks a bunch @matthdsm! :beers: !!!

PR #47, opened 2016-01-20
SSL for nginx, safer supervisord task, VIRTUAL_ENV + other env vars for Galaxy handler

pvanheus:
* Add support for nginx SSL config
* Make supervisord task ignore errors when disabled absent services
* Add VIRTUAL_ENV to galaxy supervisord config
* Add galaxy_handler_env var to allow extra environment vars for the handler

### DIFF
--- a/tasks/supervisor.yml
+++ b/tasks/supervisor.yml
@@ -18,18 +18,21 @@
   with_items:
     - uwsgi
   when: galaxy_extras_config_uwsgi and supervisor_manage_uwsgi
+  ignore_errors: yes # ignore error that occurs when service is not installed
 
 - name: Stop and remove munge.
   service: name={{ item }} state=stopped enabled=no
   with_items:
     - munge
   when: galaxy_extras_config_slurm and supervisor_manage_slurm
+  ignore_errors: yes # ignore error that occurs when service is not installed
 
 - name: Stop and remove slurm.
   service: name={{ item }} state=stopped enabled=no
   with_items:
     - slurm-llnl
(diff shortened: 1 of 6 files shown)


jmchilton:
:+1:

bgruening:
I'm a little bit worried about the ignore_error changes, because for other with this service installed, it might be useful. Could be use something like `failed_when` and filter out your specific error and only hide this one?

tasks/supervisor.yml
```diff
@@ -37,24 +40,28 @@
...
     - postgresql
   when: supervisor_manage_postgres
+  ignore_errors: yes # ignore error that occurs when service is not installed
 
 - name: Stop and remove proftpd.
   service: name={{ item }} state=stopped enabled=no
   with_items:
     - proftpd
   when: galaxy_extras_config_proftpd and supervisor_manage_proftp
+  ignore_errors: yes # ignore error that occurs when service is not installed
```
mvdbeek: @pvanheus would it be sufficient to check if galaxy_extras_install packages was yes?
when: galaxy_extras_config_proftpd and supervisor_manage_proftp and galaxy_extras_install_packages. Or alternatively, might there be a ` | bool` required to get proper evaluation of `galaxy_extras_config_proftpd` ?

pvanheus:
I think this PR needs to be split into its component parts and resubmitted.

mvdbeek:
I'll go ahead and close this PR, I think #103 covers SSL.
@pvanheus Feel free to open another PR or issue if there are other problems.

PR #48, opened 2016-01-20
Switch galaxy_root to galaxy_server_dir to bring in line ansible-galaxy.

jmchilton:
This will conflict #47 so that should be closed out first.


bgruening:
Great, thanks John!

PR #50, opened 2016-02-03
Configurable nginx proxy prefix

mvdbeek:

PR #51, opened 2016-02-19
Prefix upload related things.

mvdbeek:
Follow up to #50.
This is needed to get uploads to work when galaxy is not running at the root location.


bgruening:
Nice catch!

jmchilton:
Thanks @mvdbeek!

PR #53, opened 2016-02-21
Proxy interactive environments

mvdbeek:
With these changes port 8800 doesn't need to be exposed anymore in the docker image if you just want to server jupyter/ipython. Doesn't work for RStudio unfortunately.

### DIFF
--- a/defaults/main.yml
+++ b/defaults/main.yml
@@ -30,6 +30,10 @@ galaxy_extras_ie_jupyter_image: bgruening/docker-jupyter-notebook:16.01
 galaxy_extras_ie_fetch_rstudio: false
 galaxy_extras_ie_rstudio_image: erasche/docker-rstudio-notebook:15.10
 
+# The storage backend to use for docker-in-docker.
+# aufs on parent docker cannot be combined with aufs in child docker
+galaxy_extras_docker_storage_backend: aufs
+
 galaxy_docker_enabled: false
 galaxy_docker_sudo: false
 galaxy_docker_volumes_from: galaxy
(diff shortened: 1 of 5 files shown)


jmchilton:
Looks great. If I recall correctly this wouldn't work with RStudio - (proxying the proxy didn't work for some reason) - but nothing about these new routes prevents still exposing 8800 - so I am an enthusiastic :+1:.

Ping @bgruening, @erasche.

mvdbeek:
I'll give it a try with RStudio, I had only been testing jupyter ...

mvdbeek:
Hmm, indeed, that's too bad :/. But yeah, it shouldn't block direct access to 8800.

erasche:
I'm not going to -1 this, RStudio never got quite the love that Jupyter did (though if you really want R people, RStudio support is an absolute must).

I'm hoping some of this will be fixed with the new proxy, so I'm fine with temporarily breaking RStudio support.

mvdbeek:
Hmm, yeah, I was looking forward to using RStudio/ggplot2 myself actually.
Well, I guess I'll see how far I can get with Jupyter's R kernel.

jmchilton:
@erasche  I don't think this breaks RStudio support - I just doesn't provide the nice new feature of not requiring a proxy that Jupyter will have.

erasche:
@jmchilton ah, perhaps I misunderstood.

mvdbeek:
> @erasche I don't think this breaks RStudio support - I just doesn't provide the nice new feature of not requiring a proxy that Jupyter will have.

exactly. for RStudio it should continue to work if you also expose the node proxy port (8800 in that case).

bgruening:
:+1: works for me and as already said, it should not harm RStudio.

PR #54, opened 2016-02-27
Enable role testing with the Galaxy Docker image

bgruening:
As promised this PR will test the ansible role if it does not break the Docker Galaxy Image.

Testing now is a small chain of trust.

* bioblend has it's own unit-tests which tests a lot of functionality of a running Galaxy Server
* Galaxy Docker is using bioblend to run all tests against the Docker Image, ensuring a working Docker Image (which consumes a stable version of `ansible-galaxy-extras`)
* `ansible-galaxy-extras` is using a stable version of the Galaxy Docker Image and replaces the fixed stable `ansible-galaxy-extras`-role with a development version which need to be tested

This should ensure that all projects will work and are compatible. I hope it not to complex.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,6 +9,15 @@ install:
   # Add ansible.cfg to pick up roles path.
   - printf '[defaults]\nroles_path = ../' > ansible.cfg
 
+  # download the Galaxy Docker image and build it to test the playrole
+  - mkdir $HOME/galaxy-docker
+  - wget -q -O - https://github.com/bgruening/docker-galaxy-stable/archive/master.tar.gz | tar xzf - --strip-components=1 -C $HOME/galaxy-docker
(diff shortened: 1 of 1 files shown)


jmchilton:
Wonderful, wonderful, wonderful! This just looks like it builds the Docker image - does that Dockerfile run the tests as part of this also?

I like bioblend and I think I see why you are using it to test Galaxy, but it would be nice if Galaxy's own tests could also be run - some subset of the API tests and all of the framework tests for instance at first.

bgruening:
We could do this. My initial idea here was that we can trust Galaxy stable to run all tests, we trust Galaxy-Docker stable to run all tests, hence we can simply build the container and everything should work. 
If there is interest and no concerns about long running time we can run the framework tests as well.

PR #57, opened 2016-03-22
Adding ability to place password infront of nginx proxy

kellrott:
Additional optional config to add a basic_auth in front of nginx server. In cases, like the planemo-machine, authentication (at the galaxy level) is done automatically. This puts a small barrier to entry in front of the UI elements (the /api path gets a pass)

### DIFF
--- a/defaults/main.yml
+++ b/defaults/main.yml
@@ -67,6 +67,11 @@ nginx_conf_path: /etc/nginx/nginx.conf
 nginx_conf_directory: /etc/nginx/conf.d
 nginx_upload_store_path: "/tmp/nginx_upload_store"
 
+#web security
+nginx_use_passwords: False
+nginx_htpasswds:
+  - "admin:WiBKbsJTSQ8dc"
+
 ## Development Box Configuration
 # Configure nginx to setup a proxy to an IDE.
 nginx_proxy_ide: false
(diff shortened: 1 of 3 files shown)


defaults/main.yml
```diff
@@ -67,6 +67,11 @@ nginx_conf_path: /etc/nginx/nginx.conf
 nginx_conf_directory: /etc/nginx/conf.d
 nginx_upload_store_path: "/tmp/nginx_upload_store"
 
+#web security
+nginx_use_passwords: False
+nginx_htpasswds:
+  - "admin:WiBKbsJTSQ8dc"
+
```
jmchilton: I like everything but making it the default - can you just document here and move the updated defaults into downstream into your work?

kellrott: By default it should be off (nginx_use_passwords: False). 
Are you referring to something else?

jmchilton: Brain fart, sorry this looks great.

kellrott:
Anything else needed to get this merged?

jmchilton:
No this is perfect thanks. Can you rebase for a clean merge? Otherwise I can.

kellrott:
If it will take you less then 10 minutes, then you should do it. Otherwise, I'll go read a quick re-basing 'howto'.

jmchilton:
Merged directly as https://github.com/galaxyproject/ansible-galaxy-extras/commit/23a1012fc4f6f591ad67f437b90c7db1c9c5f94c

jmchilton:
Thanks again - this is great!

PR #59, opened 2016-04-01
Allow sharing nodes and set nativeSpecification for slurm.

mvdbeek:
The default settings up to now result in GALAXY_SLOTS=1. With this PR users can
control the nativeSpecification parameter in the job_conf.xml using the
environmental variable NATIVE_SPEC.

We also set Shared=YES in the configure_clurm.py script. Shared=YES does not change the default
(https://computing.llnl.gov/linux/slurm/cons_res_share.html) unless --share
is provided to srun or the nativeSpecification in job_conf.xml.

### DIFF
--- a/templates/configure_slurm.py.j2
+++ b/templates/configure_slurm.py.j2
@@ -77,7 +77,7 @@ SlurmctldDebug=3
 SlurmdDebug=3
 #SlurmdLogFile=
 NodeName=$hostname CPUs=$cpus RealMemory=$memory State=UNKNOWN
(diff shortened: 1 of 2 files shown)


bgruening:
:+1: Great!

jmchilton:
Agreed with @bgruening - very nice!

mvdbeek:
Thanks @bgruening @jmchilton.
What would be awesome is if we could also set the N in `--ntasks=N` based on the processor count (as returned by `nproc`, for example) when starting the handlers. Any idea how we could do this in a elegant way?

PR #60, opened 2016-04-01
Enable a bare Galaxy installation

bgruening:
This will enable Galaxy flavours with no classical Galaxy tool included by default.

### DIFF
--- a/templates/startup.sh.j2
+++ b/templates/startup.sh.j2
@@ -21,6 +21,13 @@ if [ "x$ENABLE_TTS_INSTALL" != "x" ]
         export GALAXY_CONFIG_TOOL_SHEDS_CONFIG_FILE=$GALAXY_HOME/tool_sheds_conf.xml
 fi
 
+# Remove all default tools from Galaxy by default
+if [ "x$BARE" != "x" ]
+    then
+        echo "Remove all tools from the tool_conf.xml file."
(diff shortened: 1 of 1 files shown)


templates/startup.sh.j2
```diff
@@ -21,6 +21,13 @@ if [ "x$ENABLE_TTS_INSTALL" != "x" ]
         export GALAXY_CONFIG_TOOL_SHEDS_CONFIG_FILE=$GALAXY_HOME/tool_sheds_conf.xml
 fi
 
+# Remove all default tools from Galaxy by default
+if [ "x$BARE" != "x" ]
+    then
+        echo "Remove all tools from the tool_conf.xml file."
+        export GALAXY_CONFIG_TOOL_CONFIG_FILE=config/shed_tool_conf.xml,$GALAXY_HOME/bare_tool_conf.xml
```
jmchilton: Where is ``bare_tool_conf.xml`` coming from?

bgruening: From a `ADD` statement in a Dockerfile :)

bgruening:
Thanks John.

templates/startup.sh.j2
jmchilton: You are your Docker thing :laughing:

bgruening: I take this personally ;-)

Issue #63, opened 2016-04-24
Conflict for PYTHONHOME with conda


mvdbeek:
In a fairly standard supervisor setup we get the following handler section
```
[program:handler]
command         = /home/galaxy/galaxy/.venv/bin/python ./lib/galaxy/main.py -c /home/galaxy/galaxy/config/galaxy.ini --server-name=handler%(process_num)s --log-file=/home/galaxy/galaxy/handler%(process_num)s.log
directory       = /home/galaxy/galaxy
process_name    = handler%(process_num)s
numprocs        = 1
umask           = 022
autostart       = true
autorestart     = true
startsecs       = 20
user            = galaxy
environment     = PYTHON_EGG_CACHE=/home/galaxy/.python-eggs, PYTHONHOME=/home/galaxy/galaxy.venv
startretries    = 15`
```

PYTHONHOME is in conflict with conda, so when trying to resolve a conda module we get:
```
Traceback (most recent call last):
  File "/home/galaxy/conda/bin/conda", line 3, in <module>
    from conda.cli import main
ImportError: No module named conda.cli
Traceback (most recent call last):
  File "/home/galaxy/conda/bin/conda", line 3, in <module>
    from conda.cli import main
ImportError: No module named conda.cli
Traceback (most recent call last):
  File "/home/galaxy/conda/bin/conda", line 3, in <module>
    from conda.cli import main
ImportError: No module named conda.cli
Traceback (most recent call last):
  File "/home/galaxy/conda/bin/conda", line 3, in <module>
    from conda.cli import main
ImportError: No module named conda.cli
Traceback (most recent call last):
  File "/home/galaxy/conda/bin/conda", line 3, in <module>
    from conda.cli import main
ImportError: No module named conda.cli
Traceback (most recent call last):
  File "/home/galaxy/conda/bin/conda", line 3, in <module>
    from conda.cli import main
ImportError: No module named conda.cli
```

When removing PYTHONHOME we get
```
Traceback (most recent call last):
  File "/home/galaxy/galaxy/database/jobs/000/15/set_metadata_Smnwjn.py", line 1, in <module>
    from galaxy_ext.metadata.set_metadata import set_metadata; set_metadata()
  File "/home/galaxy/galaxy/lib/galaxy_ext/metadata/set_metadata.py", line 23, in <module>
    from sqlalchemy.orm import clear_mappers
ImportError: No module named sqlalchemy.orm
```

[...]

PR #64, opened 2016-04-25
Source venv in cluster destination and remove egg references

mvdbeek:
This seems like a pretty clean way to solve #63.
(The egg removal has nothing to do with conda)

### DIFF
--- a/templates/job_conf.xml.j2
+++ b/templates/job_conf.xml.j2
@@ -33,6 +33,7 @@
   {% if galaxy_source_shellrc %}
           <env file="{{ galaxy_user_shellrc }}" />
   {% endif %}
+          <env file="{{ galaxy_venv_dir }}/bin/activate"/>
           <param id="docker_enabled" from_environ="GALAXY_DOCKER_ENABLED">{{ galaxy_docker_enabled }}</param>
           <param id="docker_sudo" from_environ="GALAXY_DOCKER_SUDO">{{ galaxy_docker_sudo }}</param>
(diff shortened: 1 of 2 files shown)


templates/job_conf.xml.j2
```diff
@@ -33,6 +33,7 @@
   {% if galaxy_source_shellrc %}
           <env file="{{ galaxy_user_shellrc }}" />
   {% endif %}
+          <env file="{{ galaxy_venv_dir }}/bin/activate"/>
```
bgruening: I can understand why this fixes it, but I'm not sure why this is needed at first hand :(
Is this now necessary for all installations?

mvdbeek: With external job handlers? Yes. But it does obviate the need for modifying PYTHONHOME for the handlers.

bgruening: Just thinking if this need to be included in galaxy main as well.

mvdbeek: Maybe ... or it may already be the case:
https://github.com/galaxyproject/usegalaxy-playbook/blob/ae79ac819ffb86a927e0a15c30f65acb08f6d3fc/templates/galaxy/test.galaxyproject.org/config/job_conf.xml.j2#L147

bgruening:
Let's merge it as it fixes this error. I'm still confused by these environment specific errors and it feels very complicated.

mvdbeek:
Thanks @bgruening, I think this is a nice step forwards towards conda in production. Do you already have some experience with this? I'm setting up our backup server ...

bgruening:
For us conda is working in produciton but our instance it too old and is not using ansible or Docker :(

PR #65, opened 2016-05-01
WIP: add HTCondor support

bgruening:
In addition this PR introduces a macro file.

The PR is completely untested. Early feedback welcome.

Depends on https://github.com/galaxyproject/galaxy/pull/2278

### DIFF
--- a/templates/job_conf.xml.j2
+++ b/templates/job_conf.xml.j2
@@ -32,14 +32,14 @@
   {% if galaxy_extras_config_pbs %}
         <destination id="pbs_cluster" runner="pbs">
             <env file="{{ galaxy_venv_dir }}/bin/activate"/>
-            {{macros.docker_enabled()}}
+            {{ macros.docker_enabled() }}
         </destination>
   {% endif %}
   {% if galaxy_extras_config_slurm %}
         <destination id="slurm_cluster" runner="slurm">
             <param id="nativeSpecification" from_environ="NATIVE_SPEC">--ntasks=1 --share</param>
             <env file="{{ galaxy_venv_dir }}/bin/activate"/>
-            {{macros.docker_enabled()}}
+            {{ macros.docker_enabled() }}
         </destination>
   {% endif %}
   {% if galaxy_extras_config_condor %}
--- a/templates/job_conf.xml.j2
+++ b/templates/job_conf.xml.j2
@@ -2,19 +2,17 @@
 {% import "macros.xml.j2" as macros with context %}
 <job_conf>
     <plugins workers="2">
-{% if galaxy_extras_config_slurm == True %}
+{% if galaxy_extras_config_slurm %}
         <plugin id="slurm" type="runner" load="galaxy.jobs.runners.slurm:SlurmJobRunner">
             <param id="drmaa_library_path">/usr/lib/slurm-drmaa/lib/libdrmaa.so</param>
         </plugin>
 {% endif %}
-{% if galaxy_extras_config_condor == True %}
-        <plugins>
-            <plugin id="condor" type="runner" load="galaxy.jobs.runners.condor:CondorJobRunner" />
-        </plugins>
+{% if galaxy_extras_config_condor %}
+        <plugin id="condor" type="runner" load="galaxy.jobs.runners.condor:CondorJobRunner" />
 {% endif %}
-{% if galaxy_extras_config_pbs == True %}
+{% if galaxy_extras_config_pbs %}
         <plugin id="pbs" type="runner" load="galaxy.jobs.runners.drmaa:DRMAAJobRunner">
-          <param id="drmaa_library_path">/usr/lib/pbs-drmaa/lib/libdrmaa.so.1</param>
+            <param id="drmaa_library_path">/usr/lib/pbs-drmaa/lib/libdrmaa.so.1</param>
         </plugin>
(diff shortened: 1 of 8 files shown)


bgruening:
@jmchilton @mvdbeek I think this PR has nice things in it that can be used even if the Galaxy PR was not merged and it should not break anything else.

Could I get a review? Thanks!

templates/job_conf.xml.j2
```diff
@@ -1,18 +1,23 @@
...
+{% if galaxy_extras_config_pbs %}
         <plugin id="pbs" type="runner" load="galaxy.jobs.runners.drmaa:DRMAAJobRunner">
-          <param id="drmaa_library_path">/usr/lib/pbs-drmaa/lib/libdrmaa.so.1</param>
+            <param id="drmaa_library_path">/usr/lib/pbs-drmaa/lib/libdrmaa.so.1</param>
         </plugin>
 {% endif %}
     </plugins>
-    <handlers default="handlers">
+    <!-- The default handler can be changed by specifying the GALAXY_HANDLERS_DEFAULT environment variable. -->
+    <handlers default_from_environ="GALAXY_HANDLERS_DEFAULT" default="handlers">
```
mvdbeek: nice!

templates/job_conf.xml.j2
```diff
@@ -21,26 +26,31 @@
...
     </handlers>
-    <destinations default="cluster">
-{% if galaxy_extras_config_slurm == True or galaxy_extras_config_pbs == True %}
-  {% if galaxy_extras_config_pbs == True %}
-      <destination id="cluster" runner="pbs">
-  {% elif galaxy_extras_config_slurm == True %}
-      <destination id="cluster" runner="slurm">
-          <param id="nativeSpecification" from_environ="NATIVE_SPEC">--ntasks=1 --share</param>
+    <!-- The default destination can be changed by specifying the GALAXY_DESTINATIONS_DEFAULT environment variable. -->
+    <destinations default_from_environ="GALAXY_DESTINATIONS_DEFAULT" default="slurm_cluster">
```
mvdbeek: Can we keep the default name "cluster"? Otherwise this breaks everything non-slurm, right?
Or alternatively can we introduce a {{ galaxy_extras_default_cluster_id }} variable?

templates/job_conf.xml.j2
```diff
@@ -21,26 +26,31 @@
...
+        </destination>
   {% endif %}
-
-  {% if galaxy_source_shellrc %}
-          <env file="{{ galaxy_user_shellrc }}" />
+  {% if galaxy_extras_config_slurm %}
+        <destination id="slurm_cluster" runner="slurm">
+            <param id="nativeSpecification" from_environ="NATIVE_SPEC">--ntasks=1 --share</param>
+            <env file="{{ galaxy_venv_dir }}/bin/activate"/>
+            {{macros.docker_enabled()}}
```
mvdbeek: lets go with the `{{ var }}` syntax. Also, where is this variable defined?

mvdbeek:
@bgruening Yep, looks nice, but I haven't tested the HTCondor part. If you can just think of something for replacing the `default="slurm_cluster">` you got the 👍

templates/job_conf.xml.j2
bgruening: Not yet working as we depend on the Galaxy patch: galaxyproject/galaxy#2278

templates/job_conf.xml.j2
bgruening: You mean the spaces? I will do! Sorry not so familiar with style guideline in ansible.

templates/job_conf.xml.j2
bgruening: As you like. I kind of like to make the naming more explicit. But yes it breaks ... {{ galaxy_extras_default_cluster_id }}

templates/job_conf.xml.j2
mvdbeek: yup. well, totally cosmetic, but for consistencies' sake it would be nice.

templates/job_conf.xml.j2
mvdbeek: but it should still work as is, just that you can't use the GALAXY_HANDLERS_DEFAULT env var, right?

bgruening: Jupp. Galaxy will just ignore this attribute.

templates/job_conf.xml.j2
mvdbeek: > I kind of like to make the naming more explicit. But yes it breaks ... {{ galaxy_extras_default_cluster_id }}

Then lets put `galaxy_extras_galaxy_destinations_default: slurm_cluster` in defaults/main.yml, and change this line to 
```
 <destinations default_from_environ="GALAXY_DESTINATIONS_DEFAULT" default="{{ galaxy_extras_galaxy_destinations_default }}">
```

bgruening: Fixed it by reverting back to cluster. This makes more sense as we only have one for the moment activated by default. Is this ok. Thanks!

mvdbeek:
Okay, this is very nice, 👍

mvdbeek:
Thanks a lot @bgruening

bgruening:
Awesome! Thanks for the merge!

PR #67, opened 2016-05-13
add galaxy_destination_default

bgruening:


bgruening:
@mvdbeek back to the default idea :)
Obviously there was a destination and the default value was not fitting the destination name. This should fix it.

mvdbeek:
right, I should have seen it. oh well. 👍

templates/job_conf.xml.j2
```diff
@@ -36,7 +36,7 @@
         </destination>
   {% endif %}
   {% if galaxy_extras_config_slurm %}
-        <destination id="slurm_cluster" runner="slurm">
+        <destination id="{{ galaxy_extras_galaxy_destination_default }}" runner="slurm">
```
mvdbeek: wait, shouldn't this be hardcoded to cluster or slurm_cluster?

bgruening: Sorry, I do not see it. The default value should be the same as the destination-id, isn't it?

mvdbeek: No, i think the destination ids should be static, and you select the default one with `{{ galaxy_extras_galaxy_destination_default }}` (or the corresponding env var).

So if you have 
```
<destination id="cluster_slurm" runner="slurm">
    <param id="nativeSpecification" from_environ="NATIVE_SPEC">--ntasks=1 --share</param>
    <env file="{{ galaxy_venv_dir }}/bin/activate"/>
    {{ macros.docker_enabled() }}
</destination>
<destination id="condor_cluster" runner="condor">
    <param id="universe" from_environ="GALAXY_CONDA_UNIVERSE">vanilla</param>
    <env file="{{ galaxy_venv_dir }}/bin/activate"/>
</destination>
```
in your job_conf, you would select slurm as the default using
`galaxy_extras_galaxy_destination_default: cluster_slurm`

bgruening:
With a big screen is makes much more sense than from a mobile :)

mvdbeek:
thanks @bgruening

PR #69, opened 2016-06-15
Add support for PBKD2 to proftpd

mvdbeek:
and allow configuration of ftp port. Thanks to @fabiorjvieira for the PKDF2
config.

### DIFF
--- a/defaults/main.yml
+++ b/defaults/main.yml
@@ -130,8 +130,14 @@ proftpd_conf_path:  /etc/proftpd/proftpd.conf
 proftpd_db_connection: galaxy@localhost
 proftpd_db_username: galaxy
 proftpd_db_password: galaxy
+# Set proftpd_sql_auth_type to SHA-1 if use_pbkd2=False in your galaxy.ini,
+# or PBKDF2 if you have use_pbkd2=True or if you haven't changed the default.
+# If you want to use PBKDF-2 you will need proftpd version 1.3.5rc3 or later
+# (which is not available by default for ubuntu-12.04)
+proftpd_sql_auth_type: SHA-1
 proftpd_welcome: "Public Galaxy FTP"
 proftpd_files_dir: /export/galaxy-central/database/ftp
+proftpd_ftp_port: 21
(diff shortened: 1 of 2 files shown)


bgruening:
@mvdbeek afaik PKDF2 is only supported for `1.3.5rc3 and later`. So it is not working with ubuntu 12.04. As this is a common platform should be issue a warning and add a `ansible_distribution_version > "12.04"`?

mvdbeek:
Fair point, how about I add a comment in defaults/main.yml ?
Since this is not the default setting, people looking for this option should see the comment,
and we don't further complicate the task.
After all proftpd may have been installed through a PPA or from source.
Or do you plan on making this the default setting?

bgruening:
A is enough. I would love to make this a default, but the Docker container will stick to ubuntu 12.04 for a long time :(
Thanks Marius!

mvdbeek:
> I would love to make this a default, but the Docker container will stick to ubuntu 12.04 for a long time :(

We should have community access to the galaxyproject-ppa, then we could make our own backport for proftpd. Let's see about this at the GCC.

bgruening:
:+1: this looks nice! Thanks @mvdbeek and @fabiorjvieira!

defaults/main.yml
```diff
@@ -130,8 +130,14 @@ proftpd_conf_path:  /etc/proftpd/proftpd.conf
 proftpd_db_connection: galaxy@localhost
 proftpd_db_username: galaxy
 proftpd_db_password: galaxy
+# Set proftpd_sql_auth_type to SHA-1 if use_pbkd2=False in your galaxy.ini,
+# or PBKDF2 if you have use_pbkd2=True or if you haven't changed the default.
+# If you want to use PBKDF-2 you will need proftpd version 1.3.5rc3 or later
+# (which is not available by default for ubuntu-12.04)
+proftpd_sql_auth_type: SHA-1
```
mvdbeek: @bgruening @afgane sorry, my bad about this, this has to be SHA1!

PR #75, opened 2016-06-26
Add routes necessary for Neo4j IE

pvanheus:


pvanheus:
The other part of https://github.com/galaxyproject/ansible-galaxy-extras/pull/74

bgruening:
@pvanheus can we add neo4j at first to Galaxy? Otherwise this might be confusing for users.

abretaud:
I *think* we can close this with all the GIE/GIT development since this was open. Feel free to reopen if it's still needed

PR #76, opened 2016-07-03
This is needed for the Docker compose work

bgruening:
ping @mvdbeek
Any reasons to not alias localhost?

### DIFF
--- a/templates/startup.sh.j2
+++ b/templates/startup.sh.j2
@@ -1,5 +1,9 @@
 #!/bin/bash
 
+# This is needed for Docker compose to have a unified alias for the main container.
+# Modifying /etc/hosts can only happen during runtime not during build-time
+echo "127.0.0.1      galaxy" >> /etc/hosts
+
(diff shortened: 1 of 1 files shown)


templates/startup.sh.j2
```diff
@@ -1,5 +1,9 @@
 #!/bin/bash
 
+# This is needed for Docker compose to have a unified alias for the main container.
+# Modifying /etc/hosts can only happen during runtime not during build-time
+echo "127.0.0.1      galaxy" >> /etc/hosts
```
mvdbeek: I would say this doesn't feel like the right way, normally you can do this either with the `--add-host` option to docker run (I think the syntax is --add-host=galaxy:127.0.0.1), or you can just give that container a hostname (`-h galaxy`).

bgruening: Yes, but this means that this container is not useable if you don't provide these options.

mvdbeek:
Well, it won't break things, so 👍

Issue #81, opened 2016-07-27
booleans not evaluated correctly in _when_ statements 


pcm32:
If there is an _and_ operator inside a _when_ statement, then booleans apparently get evaluated as strings. This gets reflected when setting the `galaxy_extras_config_slurm` and `supervisor_manage_slurm` both to false and running with the default `provision.yml` used in bgruening/docker-galaxy-stable:

```ansible
TASK [galaxy-extras : Stop and remove munge.] **********************************
failed: [localhost] (item=munge) => {"failed": true, "item": "munge", "msg": "no service or tool found for: munge"}
	to retry, use: --limit @provision.retry

PLAY RECAP *********************************************************************
localhost                  : ok=19   changed=2    unreachable=0    failed=1 
```

This is produced by:

```yaml
- name: Stop and remove munge.
  service: name={{ item }} state=stopped enabled=no
  with_items:
    - munge
  when: galaxy_extras_config_slurm and supervisor_manage_slurm
```

the undersired effect (wrong evaluation of the complete boolean construct) is only seen when you have at least one of the two variables as false, as true/true seem to evaluate as defined or filled strings, and hence true. Unfortunately, there is little documentation on the ansible site about this. This is solved by using a `|bool` filter after each boolean in the _when_ statement.

This is when running on Ubuntu 14.04 and Ansible 2.1, as installed in bgruening/docker-galaxy-stable. I have tested the suggested changes, and the final outcome is as expected (no Ansible failure and installation according to booleans set). and will submit a PR soon.

bgruening:
Wow, this is unexpected. Is this a wider ansible bug? Is this something broken in ansible maybe during the 2.0 switch?
Thanks for tracking this down!

pcm32:
No idea to be honest, I only use ansible sparingly. Could be a 2.0 break. I also see unexpected behaviour with booleans in the template files, I'll add them to the branch of the PR #82.

bgruening:
@pcm32 I think we can close this, correct?

PR #82, opened 2016-07-27
Hotfix: adds bool filters to booleans in when statements with and operators

pcm32:
This fixes issue #81 (explained in detail there). Simply adds `|bool` filters to boolean variables in _when_ statements when the statement has an _and_ operator between the variables.

### DIFF
--- a/tasks/slurm.yml
+++ b/tasks/slurm.yml
@@ -17,13 +17,13 @@
   apt: pkg={{ item }} state={{ galaxy_extras_apt_package_state }}
   with_items:
     - slurm-llnl-torque 
-  when: galaxy_extras_install_packages and ansible_distribution_version <= "15.04"
+  when: galaxy_extras_install_packages|bool and ansible_distribution_version <= "15.04"
 
 - name: Install SLURM system packages
   apt: pkg={{ item }} state={{ galaxy_extras_apt_package_state }}
   with_items:
     - slurm-wlm-torque 
-  when: galaxy_extras_install_packages and ansible_distribution_version >= "15.04"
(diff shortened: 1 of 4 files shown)


pcm32:
I see similar issues in the template files. In the startup file template, many entries get evaluated as true when in fact the boolean value is false (producing a startup with undesired entries). Adding again the `|bool` seems to work, but I presume this might be happening as well on other templates.

bgruening:
@natefoo have you seen this one? I feel this is an ansible bug and it makes everything look so ugly. I'm fine with merging it and I'm glad @pcm32 have found it.

pcm32:
Please bear with me, have found more things to correct related to this that should go on the same merge. Will change to WIP.

pcm32:
Apparently this is how you should write booleans in whens, see this [issue comment](https://github.com/ansible/ansible/issues/15636#issuecomment-215286808) on ansible.

pcm32:
Hi @bgruening I'm done with this PR, I have other improvements to add, but those will go on a separate PR. Feel free to merge this if you find it suitable, otherwise let me know. I tested all of this and seems to work fine.

bgruening:
Tested it as well and it works as far as I can see. I would like to let someone else chime in, but if no one else respond I'm happy to merge this tomorrow.

Thanks a lot for fixing this and all your other work!

afgane:
I've encountered this as well, way back with Ansible 1.3 even I think. We ended up doing the same (e.g., https://github.com/galaxyproject/ansible-cloudman-image/blob/master/tasks/main.yml) so I think this is just expected syntax as explained in the comment linked above.

bgruening:
Ok! Thanks @afgane! I realised I'm just an ansible noob.

PR #83, opened 2016-07-27
[WIP] setup configuration for proftpd to use sftp

zfrenchee:


zfrenchee:
Is there a way to change settings in `defaults/main.yml` from outside that file?

zfrenchee:
I think this is just about everything needed to make it go. How do I test this? (I've made these changes on my personal galaxy and it works, but I'd like to test it here too, if possible.

PR #86, opened 2016-07-28
[WIP] Globus (GridFTP) Integration

zfrenchee:
Not ready yet, just putting this out there so you can watch it evolve and mention any mistakes you see.


afgane:
This will be great! Please take some time to add docs for this as well, ideally for the admin as well as the end user sides.

zfrenchee:
@afgane where would the documentation go?

afgane:
I'd think that either within the main README (which needs much more work and this could be the first step in that direction) or adding another README under the `tasks` folder, which could have a section for each group of tasks then. I kind of like the second option better. Preference from someone else?

bgruening:
Detailed README in `tasks` but a link from the main README would be nice.

zfrenchee:
@bgruening @afgane @martenson @jmchilton I'd love some help with this PR -- specifically some ideas about how to structure it. Right now, I use a vars_prompt keyword, but I don't think it belongs at the bottom of the task I'm defining. Is there a better way of getting these variables at runtime? 
@bgruening how do we do this for the docker image? In the config file I have `Server = {{ ansible_fqdn }}` but I'm not sure the `ansible_fqdn` variable makes sense when we're building the image.

afgane:
Yeah, we can't bundle the user info (or the hostname) into the image since each user will want to supply their own info after they've launched the image. I think the solution here will be a boot-time hook and some documentation. The bootstrap hook will set the hostname automatically (this can be something like a systemd or startup script) while the user credentials can be obtained from  environment variables.

bgruening:
@zfrenchee such things should be configured during startup. Either in the startup.sh script or with some environment variables. You can find some examples already in the startup script.

zfrenchee:
@bgruening does startup.sh have access to the `GALAXY_CONFIG_*` flags?

bgruening:
Sure. Is has access to everything at runtime.

zfrenchee:
Also moving this over to a PR against Bjoern's docker-galaxy-stable repo.

PR #87, opened 2016-07-28
Improvements for startup and export_files jinja templates

pcm32:
- Introduces two ansible variables, `use_slurm` and `use_docker` (and uses them), as there are places in the templates that are executed regardless of whether `galaxy_extras_config_[slurm|docker]` and `supervisor_manage_[slurm|docker]` were true or false. In our use case we don't need neither slurm nor docker. But please let me know if the role of these variables is covered well by `galaxy_extras_config..` and `supervisor_manage..` and I'll switch the parts were I added control by the new variables to the original ones.
- Fixes a border case in which a directory being evaluated was a symlink.
- Replaces hard coded `/galaxy-central` and `/etc/galaxy` by the adequate ansible variables.

### DIFF
--- a/templates/export_user_files.py.j2
+++ b/templates/export_user_files.py.j2
@@ -51,7 +51,7 @@ if __name__ == "__main__":
     for filename in os.listdir('/export/'):
         if filename.startswith('welcome'):
             export_file = os.path.join( '/export/', filename)
-            image_file = os.path.join('/etc/galaxy/web/', filename)
+            image_file = os.path.join('{{ galaxy_config_dir }}/web/', filename)
             shutil.copy(export_file, image_file)
 
     if not os.path.exists( '/export{{ galaxy_server_dir }}/' ):
@@ -65,7 +65,7 @@ if __name__ == "__main__":
     # replace Galaxy's copy with these. Use symbolic link instead of copying so
     # deployer can update and reload Galaxy and changes will be reflected.
     for config in [ 'galaxy.ini', 'job_conf.xml' ]:
-        image_config = os.path.join('/etc/galaxy/', config)
+        image_config = os.path.join('{{ galaxy_config_dir }}/', config)
         export_config = os.path.join( '/export{{ galaxy_server_dir }}/config', config )
         export_sample = export_config + ".docker_sample"
         shutil.copy(image_config, export_sample)
--- a/templates/export_user_files.py.j2
+++ b/templates/export_user_files.py.j2
@@ -26,10 +26,10 @@ def change_path( src ):
             subprocess.call( 'chown -R %s:%s %s' % ( os.environ['GALAXY_UID'], os.environ['GALAXY_GID'], dest ), shell=True )
         # if destination exists (e.g. continuing a previous session), remove source and symlink
         else:
-            if os.path.isdir( src ):
+            if os.path.isdir( src ) and not os.path.islink( src.rstrip('/') ):
                 shutil.rmtree( src )
             else:
-                os.unlink( src )
+                os.unlink( src.rstrip('/') )
             os.symlink( dest, src.rstrip('/') )
 
 
--- a/templates/export_user_files.py.j2
+++ b/templates/export_user_files.py.j2
@@ -43,7 +43,7 @@ if __name__ == "__main__":
     """
     if os.path.exists( '/export/.distribution_config/' ):
         shutil.rmtree( '/export/.distribution_config/' )
-    shutil.copytree( '/galaxy-central/config/', '/export/.distribution_config/' )
+    shutil.copytree( '{{ galaxy_server_dir }}/config/', '/export/.distribution_config/' )
 
 
     # Copy all files starting with "welcome"
@@ -54,11 +54,11 @@ if __name__ == "__main__":
             image_file = os.path.join('/etc/galaxy/web/', filename)
             shutil.copy(export_file, image_file)
 
-    if not os.path.exists( '/export/galaxy-central/' ):
-        os.makedirs("/export/galaxy-central/")
-        os.chown( "/export/galaxy-central/", int(os.environ['GALAXY_UID']), int(os.environ['GALAXY_GID']) )
+    if not os.path.exists( '/export{{ galaxy_server_dir }}/' ):
+        os.makedirs("/export{{ galaxy_server_dir }}/")
+        os.chown( "/export{{ galaxy_server_dir }}/", int(os.environ['GALAXY_UID']), int(os.environ['GALAXY_GID']) )
 
-    change_path('/galaxy-central/config/')
+    change_path('{{ galaxy_server_dir }}/config/')
 
     # copy image defaults to config/<file>.docker_sample to base derivatives on,
     # and if there is a realized version of these files in the export directory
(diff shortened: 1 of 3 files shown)


templates/export_user_files.py.j2
```diff
@@ -26,10 +26,10 @@ def change_path( src ):
             subprocess.call( 'chown -R %s:%s %s' % ( os.environ['GALAXY_UID'], os.environ['GALAXY_GID'], dest ), shell=True )
         # if destination exists (e.g. continuing a previous session), remove source and symlink
         else:
-            if os.path.isdir( src ):
+            if os.path.isdir( src ) and not os.path.islink( src.rstrip('/') ):
```
bgruening: Ah, this is unfortunate. We have these changes locally here: https://github.com/bgruening/docker-galaxy-stable/blob/dev/galaxy/export_user_files.py

Do you mind porting over the changes that we have collected over the last release.
I need to improve this release process. Sorry.

bgruening:
You are referring here to job runners, right? If so I would like to see using galaxy_extras_config_slurm and docker instead of a new variable.

Imho `galaxy_extras_config_`slurm|condor|condor_docker and so on are to control job runners.
If you want to get rid of the Docker based Interactive Environment stuff, than I think a new variable to control IEs are needed.

pcm32:
So then `galaxy_extras_config_x` overrides `supervisor_manage_x`, correct? My doubts came from places like this:

```bash
{% if supervisor_manage_slurm %}
        if [[ $NONUSE != *"slurmctld"* ]]
        then
            echo "Starting slurmctld"
            supervisorctl start slurmctld
        fi
        if [[ $NONUSE != *"slurmd"* ]]
        then
            echo "Starting slurmd"
            supervisorctl start slurmd
        fi
        supervisorctl start munge

    {% else %}
        if [[ $NONUSE != *"slurmctld"* ]]
        then
            echo "Starting slurmctld"
            /usr/sbin/slurmctld -L {{ slurm_log_dir }}/slurmctld.log
        fi
        if [[ $NONUSE != *"slurmd"* ]]
        then
            echo "Starting slurmd"
            /usr/sbin/slurmd -L {{ slurm_log_dir }}/slurmd.log
        fi

        # We need to run munged regardless
        mkdir -p /var/run/munge && /usr/sbin/munged -f
    {% endif %}
```
which seemed to me that you could have processes running (in this case, slurm), even if galaxy_config and supervisor_manage would forbid it. Would it be correct, under what you tell me, to wrap this statement completely on a `if galaxy_config_slurm`?

Also there was the `unmount` line for docker that had no boolean controlling it.

But I will change back to the existing variables, no problem!

What do you mean exactly by _Interactive Environment_ in this context? I saw some mentions to it in some files, but couldn't figure out what it meant exactly.

templates/export_user_files.py.j2
pcm32: I was wondering why you did an ADD of that file in the docker file if this was part of the ansible-extras, but didn't dig in any further than that. This explains it, hehehe.

pcm32:
Ok, I found some slides you have on the Interactive Environment... looks very neat! I think it would be useful for us to use it, but for our use case, I would like to have those containers deployed through kubernetes instead of through galaxy. So, yes, I want IE, but, no, I don't want docker, which gives a use case for this configuration to detach those two.

bgruening:
@pcm32 there are multiple problems here.
One thing is that there is a use-case that you want to configure slurm (condor etc.) and offer all options but do not start it during startup.

The other thing is that IEs currently need a docker client installed. This is because IEs do not use the job-runners yet (this is planned and part of a grant if I'm not misleading). So IEs only support local Docker, remote Docker with a local Docker-client or Docker Swarm. This is the reason we still need to have Docker in this role. 

The `unmount` is used for Docker in Docker which is needed for IEs. This can be made optional if desired.

I hope this helps a little bit to explain a few things in this role.

pcm32:
> One thing is that there is a use-case that you want to configure slurm (condor etc.) and offer all options but do not start it during startup.

For this to happen essentially you need to have the proper `galaxy_extras_config_*: true`  and `supervisor_*_manage: true` (in some cases), correct? then have those tools not being triggered in the startup script when calling it, right? For this you seem to have a different startup script (lite version) or the NONUSE variables, correct?

It seems to me that, in the current scenario, /usr/bin/startup will simply startup whatever has a `galaxy_extra_config_` on true. The `supervisor_*_manage` apparently only decides whether the execution will be controlled via supervisor (when true) or by some other method (when false, but only in some cases apparently; for instance docker only seems to start by supervisor only, right?)

> The other thing is that IEs currently need a docker client installed. This is because IEs do not use the job-runners yet (this is planned and part of a grant if I'm not misleading). So IEs only support local Docker, remote Docker with a local Docker-client or Docker Swarm. This is the reason we still need to have Docker in this role.

Ok, so, at the moment, IEs are indivisible from docker containers, understood. However, you might still want to use docker without turning on IEs, so still docker should be able to be controlled on its own. I see now that you do this through `supervisor_manage_docker`, as you configure docker through apt-gets before. On my variant (of the dockerfile) I avoid installing docker altogether. Maybe this should be incorporated into the ansible part.

I presume then that docker-within-docker is not a requirement for IE, however, unmount then shouldn't be controlled neither by galaxy_extras_configure_docker nor galaxy_extras_config_ie_proxy. Do we need a new boolean here?

```bash
{% if galaxy_extras_configure_docker_w_docker %}
umount /var/lib/docker
{% endif %}
``` 

or what would you use there? I don't mind so much about the docker umount as long as startup doesn't fail for not finding it...

bgruening:
> For this to happen essentially you need to have the proper `galaxy_extras_config_*: true`  and `supervisor_*_manage: true` (in some cases), correct? then have those tools not being triggered in the startup script when calling it, right? 

Yes.

> For this you seem to have a different startup script (lite version) or the NONUSE variables, correct?

This is more complicated. The lite version does something really different, it does not start anything besides the python webserver from Galaxy. So really lite.
The NONUSE variable is the one thats decides which processes to start.

> It seems to me that, in the current scenario, /usr/bin/startup will simply startup whatever has a `galaxy_extra_config_` on true. The `supervisor_*_manage` apparently only decides whether the execution will be controlled via supervisor (when true) or by some other method (when false, but only in some cases apparently; for instance docker only seems to start by supervisor only, right?)

Mh, seems to. Not sure who is using the non supervisor setup. Maybe we can deprecate this route and remove it at some point completely to make the roles simpler.


> Ok, so, at the moment, IEs are indivisible from docker containers, understood. However, you might still want to use docker without turning on IEs, so still docker should be able to be controlled on its own. I see now that you do this through `supervisor_manage_docker`, as you configure docker through apt-gets before. On my variant (of the dockerfile) I avoid installing docker altogether. Maybe this should be incorporated into the ansible part.

I'm fine with this.

>I presume then that docker-within-docker is not a requirement for IE, however, unmount then shouldn't be controlled neither by galaxy_extras_configure_docker nor galaxy_extras_config_ie_proxy. Do we need a new boolean here?

```bash
{% if galaxy_extras_configure_docker_w_docker %}
umount /var/lib/docker
{% endif %}
``` 

> or what would you use there? I don't mind so much about the docker umount as long as startup doesn't fail for not finding it...

Maybe we can make this bash command smarter and silent. One extra flag seems to much for me. It does have no influence at all.

Thanks @pcm32 for your work on this!
Bjoern

jmchilton:
I think planemo machine manages Docker through Ubuntu instead supervisor - which is in general a good practice I think. https://github.com/galaxyproject/planemo-machine/blob/master/group_vars/all#L89. I might be okay with simplifications - kind of depends on what they look like though.

bgruening:
But `through Ubuntu` means we need to switch from init.d scripts (or upstart) to systemd at some point. So essentially we restrict this role to a few OS versions or we add more and more complexity.

Not saying this is bad, just wanted to point out that we can simplify a lot if we agree on a common process manager or on a very limited set of supported OS versions.

bgruening:
@pcm32 some updates from my side: https://github.com/galaxyproject/ansible-galaxy-extras/pull/102

pcm32:
I don't think I ever used any of this work, although some ideas are still valid I think... I propose that we close this PR, and I can cherry pick things from here if needed later on.

PR #88, opened 2016-07-28
Https via LetsEncrypt

zfrenchee:
Another work in progress.


martenson:
this is great!

zfrenchee:
@martenson @jmchilton How does one go about testing these?

bgruening:
This looks great! Can you add some documentation about this?

zfrenchee:
@bgruening I like your enthusiasm! however...
1. I haven't tested it, and it's mostly copy-pasted, so I doubt it works. How do I test it? 
2. Where would you like for me to add documentation?

bgruening:
@zfrenchee testing is hard I think.
A few links to consider: https://docs.ansible.com/ansible/letsencrypt_module.html and look at other letsencrypt ansible tests, e.g. https://github.com/bytepark/ansible-role-letsencrypt/blob/master/.travis.yml

zfrenchee:
@bgruening Oh my gosh I can't believe I didn't see that ansible module when i first went around looking for this! Let me revise this PR.

zfrenchee:
So my guess is that nearly all the fields should be parameters passed along to the end user. @bgruening let me know what you think.

zfrenchee:
Unfortunately I don't think this will work, because running the ansible role during production of @bgruening's docker image will cause the image itself to have a certificate, which when multiple people go to run it, breaks the entire idea of https, I think. I don't know these technologies too well, but I think this is a fundamental roadblock for me, since I want to use these via bjoern's image. That said, I think others will likely be able to use this, if they are the ones to run this role, rather than use it through the docker image, so i've left it here, mostly finished, for someone else to pick up, or for myself to pick up later when I have more time. In the meantime, I'm going to look into a docker-based https solution for my work.

martenson:
@zfrenchee thank you very much for exploring this road, I still think it is great direction and that deployers would benefit from having something like this readily available when deploying Galaxy

blankenberg:
:heart: LetsEncrypt

bgruening:
@zfrenchee one possibility is to generate this key/certificate during statup via the startup script, that should make it unique to every instance.

zfrenchee:
@bgruening how would I do that?

zfrenchee:
Alright, I'm moving this work over to a PR against Bjoern's image.

PR #89, opened 2016-07-30
Add Restart Nginx handler

mvdbeek:
This should solve the recent test failures. Use capitals for the handler, since we do that in ansible-galaxy-tools as well.

Ping @bgruening


bgruening:
Thanks! :+1:

PR #90, opened 2016-08-02
remove psutil

bgruening:
psutil is nowadays in Galaxy and will be installed via Galaxy wheels. So we can remove it from this role. https://github.com/galaxyproject/galaxy/blob/431c7707b5f5d9f424267b8e059c2e2d42deb9df/lib/galaxy/dependencies/requirements.txt#L48

### DIFF
--- a/tasks/pbs.yml
+++ b/tasks/pbs.yml
@@ -3,7 +3,6 @@
   with_items:
     - torque-client
     - pbs-drmaa-dev
-    - python-psutil
   when: galaxy_extras_install_packages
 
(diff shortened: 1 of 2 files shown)


jmchilton:
I thought it was used by the slurm startup script not Galaxy?

bgruening:
Let me check the runtime environment for this script, but yes you are right. Any concerns about the pbs one?

jmchilton:
By all means remove it from the pbs list - I think that makes good sense.

jmchilton:
Thanks for the cleanup @bgruening

PR #85, opened 2016-07-27
Allow for SFTP instead of FTP

zfrenchee:
I moved this from my master branch into an actual branch and the previous (#83) PR disappeared.

PR #93, opened 2016-08-04
fix a bug in #85

zfrenchee:


tasks/proftpd.yml
```diff
@@ -24,6 +24,13 @@
     - LoadModule mod_sftp_sql.c
   when: proftpd_use_sftp
 
+- name: OpenSSH | Make sure server and client packages are installed
+  apt:
+    pkg: "{{item}}"
+    state: present
+  with_items: ["openssh-client", "openssh-server"]
```
bgruening: We don't want to have the server, client is enough I think.
xref: https://github.com/galaxyproject/ansible-galaxy-extras/pull/85/files/d693f4798621aa933e43f023ec9a44bcba498e22#r73215865 ;)

PR #96, opened 2016-08-10
Make ftp and sftp stand side by side, instead of replacing ftp with sftp

zfrenchee:
I've not yet taken the step of removing the flag entirely, which would turn on sftp and ftp on together by default. I can't really imagine anyone *not* wanting the option of sftp, but maybe it makes sense to keep the flag. Let me know @martenson @mvdbeek @bgruening @jmchilton

### DIFF
--- a/defaults/main.yml
+++ b/defaults/main.yml
@@ -150,6 +150,7 @@ proftpd_passive_port_low: 30000
 proftpd_passive_port_high: 40000
 # Set proftpd_use_sftp to true to use sftp instead of ftp
 proftpd_use_sftp: false
+proftpd_sftp_port: 8022
 # Set masquearade to true if host is NAT'ed.
 proftpd_nat_masquerade: false
 # proftpd_masquerade_address refefers to the ip that clients use to establish an ftp connection.
--- a/defaults/main.yml
+++ b/defaults/main.yml
@@ -150,7 +150,7 @@ proftpd_passive_port_low: 30000
 proftpd_passive_port_high: 40000
 # Set proftpd_use_sftp to true to use sftp instead of ftp
 proftpd_use_sftp: false
-proftpd_sftp_port: 8022
+proftpd_sftp_port: 22
 # Set masquearade to true if host is NAT'ed.
 proftpd_nat_masquerade: false
 # proftpd_masquerade_address refefers to the ip that clients use to establish an ftp connection.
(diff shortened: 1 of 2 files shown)


templates/proftpd.conf.j2
```diff
@@ -18,15 +18,16 @@ Group                           nogroup
...
 
 {% if proftpd_use_sftp %}
-SFTPEngine                      on
-
-# Configure both the RSA and DSA host keys, using the same host key files that OpenSSH uses.
-SFTPHostKey                     /etc/proftpd/ssh_host_keys/rsa
-SFTPHostKey                     /etc/proftpd/ssh_host_keys/dsa
-SFTPCompression                 delayed
+<VirtualHost 0.0.0.0>
+    Port                                    2222
```
mvdbeek: Can you make the port configurable? (`{{ proftpd_sftp_port }}`) ?

mvdbeek:
Does this work now? I was under the impression that you have to create one VirtualHost section for ftp and one for sftp, and the sftp section should be wrapped in `<IfModule mod_sftp.c>`.
(See https://serverfault.com/a/571232) but I haven't tested this yet.

zfrenchee:
I believe it works. 

They don't both need to be VirtualHosts, see the [documentation](http://www.proftpd.org/docs/contrib/mod_sftp.html#SFTPAndFTP).

I initially had `<IfModule mod_sftp.c>` but `mod_sftp.c` is included when `proftpd_use_sftp` is turned on, so in order not to use weirdly cascading booleans I decided to keep a single point of truth, which is `proftpd_use_sftp` (which I'd like to remove by the way, see my first comment).

templates/proftpd.conf.j2
zfrenchee: Done.

defaults/main.yml
```diff
@@ -150,6 +150,7 @@ proftpd_passive_port_low: 30000
 proftpd_passive_port_high: 40000
 # Set proftpd_use_sftp to true to use sftp instead of ftp
 proftpd_use_sftp: false
+proftpd_sftp_port: 8022
```
bgruening: 22, isn't it?

templates/proftpd.conf.j2
```diff
@@ -18,15 +18,16 @@ Group                           nogroup
...
 
 {% if proftpd_use_sftp %}
-SFTPEngine                      on
-
-# Configure both the RSA and DSA host keys, using the same host key files that OpenSSH uses.
-SFTPHostKey                     /etc/proftpd/ssh_host_keys/rsa
-SFTPHostKey                     /etc/proftpd/ssh_host_keys/dsa
-SFTPCompression                 delayed
+<VirtualHost 0.0.0.0>
+    Port                                    {{ proftpd_sftp_port }}
```
bgruening: Can you fix the indention of the second column?

zfrenchee: done.

bgruening:
Thanks @mvdbeek and @zfrenchee!

PR #98, opened 2016-08-11
Fix port number for SFTP config

zfrenchee:
@bgruening


zfrenchee:
Actually doesn't look like it's a port issue.

PR #102, opened 2016-08-14
update the export_user_file.py changes with the upstream one

bgruening:
The one upstream should be deleted, so that there is no confusion anymore.


bgruening:
Rebased with the `/tool_deps` changes included.
ping @jmchilton

PR #106, opened 2016-08-23
Rabbitmq

mvdbeek:
This goes with https://github.com/galaxyproject/ansible-galaxy-os/pull/20.
TL;DR with rabbitmq we can automatically reload tool data tables after data managers have run (and other things).


jmchilton:
:+1:

bgruening:
Thanks!

PR #108, opened 2016-08-25
Make sure the Galaxy log dir exists.

afgane:
Variable `galaxy_log_dir` carries a default value and it is used in a number of tasks so ensure the given dir exists.


bgruening:
Thanks @afgane

PR #110, opened 2016-08-29
Add role meta folder

afgane:
This allows the role to be installed via `ansible-galaxy` command, which can be desirable in general and needed by https://github.com/ARTbio/GalaxyKickStart/issues/186.


mvdbeek:
Nice, thanks @afgane.

PR #116, opened 2016-11-23
Tags and startup.sh

voegtlel:
Added tags for nginx, proftpd, slurm.
startup.sh always creates the admin user in case it doesn't exist.

### DIFF
--- a/templates/startup.sh.j2
+++ b/templates/startup.sh.j2
@@ -271,3 +271,9 @@ if [ "x$DISABLE_REPORTS_AUTH" != "x" ]
         echo "Disable Galaxy reports authentification "
         rm /etc/nginx/htpasswd
 fi
+
+# In case the user wants the default admin to be created, do so.
+if [ "x$GALAXY_DEFAULT_ADMIN_USER" != "x" ]
+    then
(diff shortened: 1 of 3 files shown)


templates/startup.sh.j2
```diff
@@ -271,3 +271,9 @@ if [ "x$DISABLE_REPORTS_AUTH" != "x" ]
         echo "Disable Galaxy reports authentification "
         rm /etc/nginx/htpasswd
 fi
+
+# In case the user wants the default admin to be created, do so.
+if [ "x$GALAXY_DEFAULT_ADMIN_USER" != "x" ]
+    then
+    python /usr/local/bin/create_galaxy_user.py --user $GALAXY_DEFAULT_ADMIN_USER --password $GALAXY_DEFAULT_ADMIN_PASSWORD -c $GALAXY_CONFIG_FILE --key $GALAXY_DEFAULT_ADMIN_KEY
```
mvdbeek: @voegtlel 
I would probably single-quote everything in here, so that a space or other funny things in the password don't break this.

bgruening: :+1: single quoting env sounds good!

bgruening:
@mvdbeek any recommendation on tag-names? Is this ok for you?

mvdbeek:
The more the better :). Yes, this is fine.

bgruening:
Awesome, thanks to you both!

PR #117, opened 2016-11-24
Order of commands

voegtlel:
Fixed order of commands (also for DISABLE_REPORTS_AUTH).
Added quotes as mentioned in #116 
Cleaned up create_galaxy_user.py
Added check_database.py

### DIFF
--- a/templates/create_galaxy_user.py.j2
+++ b/templates/create_galaxy_user.py.j2
@@ -3,7 +3,7 @@ import sys
 sys.path.insert(1,'{{ galaxy_server_dir }}')
 sys.path.insert(1,'{{ galaxy_server_dir }}/lib')
 
-from galaxy.model import *
+from galaxy.model import User, APIKeys
 from galaxy.model.mapping import init
 from galaxy.model.orm.scripts import get_config
 import argparse
--- a/templates/create_galaxy_user.py.j2
+++ b/templates/create_galaxy_user.py.j2
@@ -1,24 +1,14 @@
+#!/usr/bin/env python
 import sys
 sys.path.insert(1,'{{ galaxy_server_dir }}')
-sys.path.insert(1,'{{ galaxy_server_dir }}')
+sys.path.insert(1,'{{ galaxy_server_dir }}/lib')
 
-from scripts.db_shell import *
-from galaxy.util.bunch import Bunch
-from galaxy.security import GalaxyRBACAgent
-from sqlalchemy.orm import sessionmaker
-from sqlalchemy import *
+from galaxy.model import *
+from galaxy.model.mapping import init
+from galaxy.model.orm.scripts import get_config
 import argparse
-bunch = Bunch( **globals() )
-engine = create_engine('{{ galaxy_database_connection }}')
-bunch.session = sessionmaker(bind=engine)
-# For backward compatibility with "model.context.current"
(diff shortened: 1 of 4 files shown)


templates/startup.sh.j2
```diff
@@ -275,5 +275,15 @@ fi
 # In case the user wants the default admin to be created, do so.
 if [ "x$GALAXY_DEFAULT_ADMIN_USER" != "x" ]
     then
-    python /usr/local/bin/create_galaxy_user.py --user $GALAXY_DEFAULT_ADMIN_USER --password $GALAXY_DEFAULT_ADMIN_PASSWORD -c $GALAXY_CONFIG_FILE --key $GALAXY_DEFAULT_ADMIN_KEY
+        echo "Creating admin user $GALAXY_DEFAULT_ADMIN_USER with key $GALAXY_DEFAULT_ADMIN_KEY and password $GALAXY_DEFAULT_ADMIN_PASSWORD if not existing"
```
bgruening: Can you single quote the variables again.

templates/startup.sh.j2
```diff
@@ -275,5 +275,15 @@ fi
...
+        python /usr/local/bin/create_galaxy_user.py --user "$GALAXY_DEFAULT_ADMIN_USER" --password "$GALAXY_DEFAULT_ADMIN_PASSWORD" -c "$GALAXY_CONFIG_FILE" --key "$GALAXY_DEFAULT_ADMIN_KEY"
```
bgruening: Can you single quote the variables again.

templates/startup.sh.j2
```diff
@@ -117,6 +117,13 @@ then
     ln -s /export/act_qmaster /var/lib/gridengine/default/common/act_qmaster
 fi
 
+# Waits until postgres is ready
+function wait_for_postgres {
+    echo "Checking if database is up and running"
+    until /usr/local/bin/check_database.py >/dev/null; do sleep 1; echo "Waiting for database"; done
```
bgruening: Neat!

templates/create_galaxy_user.py.j2
```diff
@@ -58,7 +49,12 @@ if __name__ == "__main__":
     parser.add_argument("--key", help="API-Key.")
     parser.add_argument("--username", default="admin",
                     help="The public username. Public names must be at least three characters in length and contain only lower-case letters, numbers, and the '-' character.")
+    parser.add_argument('args', nargs=argparse.REMAINDER)
 
     options = parser.parse_args()
 
-    add_user(options.user, options.password, key=options.key, username=options.username)
+    mapping = init('/tmp/', db_url)
```
bgruening: Does this need to be `/tmp/` or can this be a python tempfile object?

mvdbeek:
Postgres comes with a pg_isready binary, any reason to not use it instead?

voegtlel:
Yes, I thought about it, but then it needs the server+user+password+database variables separately, which means they'd have to be split from
GALAXY_CONFIG_DATABASE_CONNECTION=postgresql://user:password@host/database?client_encoding=utf8
Also this approach is more universal, because it could also take another type of external database (mysql, etc).

mvdbeek:
Fair enough!

bgruening:
LGTM! Thanks

PR #118, opened 2016-11-25
Change galaxy_docker_volumes_from to ""

bgruening:
@jmchilton do you know why this was set to `galaxy` there is no data-volume called galaxy available, isn't it.

### DIFF
--- a/defaults/main.yml
+++ b/defaults/main.yml
@@ -51,7 +51,7 @@ galaxy_extras_docker_storage_backend: aufs
 
 galaxy_docker_enabled: false
 galaxy_docker_sudo: false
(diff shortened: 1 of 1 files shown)


bgruening:
ping @stevecassidy

jmchilton:
Sorry for missing this - I don't know why I always miss issues and PRs on this repo.

I think to enable Docker-in-Docker - does this break something?

bgruening:
For Docker-in-Docker this needs to be removed I think - at least this is what we figured out with the NLP flavor. I'm just wondering why this was added. This parameter means there needs to be a volume called 'galaxy' correct?

bgruening:
ping @stevecassidy

PR #121, opened 2017-01-19
Add remote_user header if received from another proxy

abretaud:
Hej,
I needed to add these to make the authentication work on my setup which looks like this:

The Internet --> nginx proxy doing auth with ldap module, filling a REMOTE_USER header --> nginx embedded in docker-galaxy-stable

### DIFF
--- a/templates/nginx_galaxy_web.conf.j2
+++ b/templates/nginx_galaxy_web.conf.j2
@@ -6,7 +6,8 @@ location {{ nginx_galaxy_location }}/ {
 {% if galaxy_admin_user is defined and galaxy_admin_user %}
     # hard-code a fixed user to pass to Galaxy to auto-login
 	proxy_set_header REMOTE_USER '{{ galaxy_admin_user }}';
-{% else %}
+{% endif %}
+{% if nginx_use_remote_header %}
     # forward the remote_user header in case it is set by a previous proxy
     if ($remote_user) {
         proxy_set_header REMOTE_USER $remote_user;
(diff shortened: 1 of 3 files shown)


templates/nginx_galaxy_web.conf.j2
```diff
@@ -2,9 +2,14 @@
 location {{ nginx_galaxy_location }}/ {
 	proxy_pass http://galaxy_web_app;
 	proxy_set_header X-Forwarded-Host $host;
-	proxy_set_header X-Forwarded-For  $proxy_add_x_forwarded_for;     
+	proxy_set_header X-Forwarded-For  $proxy_add_x_forwarded_for;
 {% if galaxy_admin_user is defined and galaxy_admin_user %}
     # hard-code a fixed user to pass to Galaxy to auto-login
 	proxy_set_header REMOTE_USER '{{ galaxy_admin_user }}';
+{% else %}
```
mvdbeek: I think it would be clearer if this could be a separate setting ("nginx_use_remote_header" maybe?).
With this PR the logic would be "If I want to pass the remote_user header, I need to unset galaxy_admin_user", which is a bit weird, no?
Other than that 👍

mvdbeek:
@abretaud sorry, should have mentioned this in my first comment, but can you also set the default (False) for this new variable somewhere around https://github.com/galaxyproject/ansible-galaxy-extras/blob/master/defaults/main.yml#L108

mvdbeek:
wow so fast! ;)

abretaud:
hehe, I benefiting from live @bgruening support :)

mvdbeek:
Alright, sorry for the delay. Thanks @abretaud !

PR #122, opened 2017-01-22
Add docker to condor

bgruening:


jmchilton:
Thanks a bunch for this?

Do you know if there is any active effort to create virtual machines demonstrating the Condor or Kubernates work?

bgruening:
We do have a project to run Galaxy with ht-condor to submit containers. And we do have a compose setup that we want to test on Kubernetes.

PR #124, opened 2017-02-20
Create /.galaxy_save file and check it.

manabuishii:
Comment describe user re-starts with docker start and stop.
But I can't find any /.galaxy_file.

So this PullRequest create /.galaxy_file after all setup finished by this script at container start first time.
And check /.galaxy_file when this script's enter main to avoid this script run twice.

### DIFF
--- a/templates/export_user_files.py.j2
+++ b/templates/export_user_files.py.j2
@@ -50,8 +50,7 @@ if __name__ == "__main__":
         is aborted.
     """
     if os.path.exists( '/.galaxy_save' )
-        print("It is already setup. So we do nothing.")
-        sys.exit(0)
+        sys.exit("It is already setup. So we do nothing.")
     galaxy_root_dir = os.environ.get('GALAXY_ROOT', '/galaxy-central/')
 
     galaxy_distrib_paths = {'/galaxy-central/config/': '/export/.distribution_config',
--- a/templates/export_user_files.py.j2
+++ b/templates/export_user_files.py.j2
@@ -49,7 +49,8 @@ if __name__ == "__main__":
         If the user re-starts (with docker start) the container the file /.galaxy_save is found and the linking
         is aborted.
     """
-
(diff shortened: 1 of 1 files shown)


templates/export_user_files.py.j2
```diff
@@ -120,3 +121,6 @@ if __name__ == "__main__":
         subprocess.call('chmod -R 0755 /export/', shell=True)
         subprocess.call('chmod -R 0700 %s' % PG_DATA_DIR_HOST, shell=True)
 
+    file = open('/.galaxy_save', 'w')
```
bgruening: You can use `with open() as` here, or `open(fname, 'a').close()
`

templates/export_user_files.py.j2
```diff
@@ -49,7 +49,8 @@ if __name__ == "__main__":
         If the user re-starts (with docker start) the container the file /.galaxy_save is found and the linking
         is aborted.
     """
-
+    if os.path.exists( '/.galaxy_save' )
+        sys.exit(0)
```
bgruening: You might want to return a meaningful message here.

bgruening:
@manabuishii can you be a little bit more descriptive in your PR description. I think I know what you mean, but not entirely sure.

templates/export_user_files.py.j2
manabuishii: I fix this with 
`open(fname, 'a').close()`
style

templates/export_user_files.py.j2
manabuishii: I add message before `exit(0)`

bgruening:
@manabuishii sorry missed your latest changes.
I'm still not sure this is needed, can you not simply use `.distribution_config`?

manabuishii:
When docker container is removed or docker daemon is down by accident, re-run container.
Container is need to setup but `.distribution_config` is exists under export directory.
At least some links will not be created .

PR #126, opened 2017-03-24
More options for compose setup

voegtlel:
Proftpd option not to regenerate ssh keys, install-repository option for specifying the target instance


bgruening:
Thanks @voegtlel!

PR #127, opened 2017-04-14
Configure proxy prefix

abretaud:
Hej!
I'm running galaxy dockers behind a collection of proxies, with url prefixes and it was a bit complicated to configure until now. So I'm preparing a PR for https://github.com/bgruening/docker-galaxy-stable where I introduce a new env var (GALAXY_CONFIG_PROXY_PREFIX) which will magically make all changes necessary to make all components aware of the url prefix.
The main logic is in this PR, most of it is working on my test setup, though I'll do more tests next week.
Open to any suggestions of course!
Happy easter! 🥚 🥚 🥚

### DIFF
--- a/templates/startup.sh.j2
+++ b/templates/startup.sh.j2
@@ -49,32 +49,45 @@ echo "127.0.0.1      galaxy" >> /etc/hosts
 ansible localhost -m ini_file -a "dest=/etc/supervisor/conf.d/galaxy.conf section=program:handler option=numprocs value=${GALAXY_HANDLER_NUMPROCS:-2}" &> /dev/null
 
 # Configure proxy prefix filtering
-if [ "$GALAXY_CONFIG_PROXY_PREFIX" != "/" ]
+if [ "$PROXY_PREFIX" != "/" ]
     then
-    ansible localhost -m ini_file -a "dest=${GALAXY_CONFIG_FILE} section=filter:proxy-prefix option=prefix value=${GALAXY_CONFIG_PROXY_PREFIX}" &> /dev/null
+    ansible localhost -m ini_file -a "dest=${GALAXY_CONFIG_FILE} section=filter:proxy-prefix option=prefix value=${PROXY_PREFIX}" &> /dev/null
     ansible localhost -m ini_file -a "dest=${GALAXY_CONFIG_FILE} section=app:main option=filter-with value=proxy-prefix" &> /dev/null
 
-    ansible localhost -m ini_file -a "dest=${GALAXY_CONFIG_DIR}/reports_wsgi.ini section=filter:proxy-prefix option=prefix value=${GALAXY_CONFIG_PROXY_PREFIX}/reports" &> /dev/null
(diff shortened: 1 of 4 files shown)


bgruening:
@abretaud great! Thanks for working on this! Can we change `GALAXY_CONFIG_PROXY_PREFIX` to something else. Usually GALAXY_CONFIG_* indicates that it is a variable from galaxy.ini. Mabye just `PROXY_PREFIX`?

abretaud:
Right, I just changed its name. From what I have tested it works ok now.

I had a problem accessing reports when DISABLE_REPORTS_AUTH was true, it looks like there was a bug in the nginx conf (because of the [strange if behavior in nginx](https://www.nginx.com/resources/wiki/start/topics/depth/ifisevil/)). I reworked it to avoid using the nginx `if`, it looks a bit hacky, but I see no better way...

I have problems running docker-in-docker at the moment, so I couldn't test IE. (Yet)

templates/startup.sh.j2
```diff
@@ -49,34 +49,47 @@ echo "127.0.0.1      galaxy" >> /etc/hosts
...
 
+# Disable authentication of Galaxy reports
+if [ "x$DISABLE_REPORTS_AUTH" != "x" ]
+    then
+        # disable authentification by deleting the htpasswd file
+        echo "Disable Galaxy reports authentification "
+        rm /etc/nginx/htpasswd
+        echo "" > /etc/nginx/conf.d/reports_auth.conf
+    else
+        # enable authentification by deleting the htpasswd file
```
bgruening: This comment looks wrong.

bgruening:
@jmchilton @mvdbeek any comment to this PR. Looks good to me.

templates/startup.sh.j2
```diff
@@ -48,6 +48,48 @@ echo "127.0.0.1      galaxy" >> /etc/hosts
...
+    fi
+fi
+
+# Disable authentication of Galaxy reports
+if [ "x$DISABLE_REPORTS_AUTH" != "x" ]
+    then
+        # disable authentification by deleting the htpasswd file
+        echo "Disable Galaxy reports authentification "
+        rm /etc/nginx/htpasswd
+        echo "" > /etc/nginx/conf.d/reports_auth.conf
```
mvdbeek: Why not keep
```
            if (!-f /etc/nginx/htpasswd) {		
                set $auth off;		
            }
```
in templates/nginx_reports_auth.conf.j2 ?
That would be better for people not using the docker image / these startup scripts.

abretaud: I'd prefer to, but it doesn't work unfortunately... When nginx enters into the if {}, the proxy_pass is ignored and you get a 404 error. I had a hard time finding out why, there are some examples explaining why there: http://agentzh.blogspot.fr/2011/03/how-nginx-location-if-works.html

bgruening:
@mvdbeek what do you think?

mvdbeek:
This does now imperatively required a htpwassd file to use reports, and the role is not providing a way to set the htpasswd file. I'd prefer this to be optional, so that you can run the majority of the role without additional bash action. 
You could even remove the bash scripting part in this PR and run ansible to apply the new settings at runtime, like we already do for letsencrypt (lines 332-360 in startup.sh.j2). 
The PR is good otherwise, and setting proxy-prefix at runtime is definitely important. If this will bite me in the future I can always open another PR.

PR #128, opened 2017-04-14
remove uwsgi restart

bgruening:
I don't think this is needed and I have problems if I only include a uwsgi from pip.

### DIFF
--- a/tasks/supervisor.yml
+++ b/tasks/supervisor.yml
@@ -14,12 +14,6 @@
   service: name=supervisor state=stopped
   tags: stop_supervisor
 
-- name: Stop and remove uwsgi.
-  service: name={{ item }} state=stopped enabled=no
-  with_items:
-    - uwsgi
-  when: galaxy_extras_config_uwsgi|bool and supervisor_manage_uwsgi|bool
-
 - name: Stop and remove munge.
   service: name={{ item }} state=stopped enabled=no
   with_items:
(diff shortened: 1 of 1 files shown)


bgruening:
@jmchilton @mvdbeek any reason to keep this?

mvdbeek:
So the idea is to disable automatic uwsgi starts if it has been installed through the package manager. This is not a problem if a) you're using docker and docker restricts automatic service starting. But in ./tasks/supervisor.yml there is this section:
```
- name: Install web related system packages
  apt: pkg={{ item }} state={{ galaxy_extras_apt_package_state }}
  with_items:
    - uwsgi
    - uwsgi-plugin-python
    - supervisor
  when: galaxy_extras_install_packages
```
Which will cause uwsgi to be installed from apt, which starts a uwsgi service on non-docker systems. If you want to remove this here you also need to remove the installation of uwsgi, or you can add an `ignore_errors: yes` on the part that causes problems, or even better add `galaxy_extras_install_packages` to the conditions for this task.

bgruening:
@mvdbeek if this is the reason, maybe we should install both via pip to get a more recent version?

mvdbeek:
That would be good as well, yes.

bgruening:
I have disabled it for now and will work on the pip stuff later.

mvdbeek:
Thanks @bgruening!

bgruening:
Cool, thanks for the merge. Next release we will have a `s17.05` version, that is a down-scaled version of Galaxy Docker with 150MB less image size: https://quay.io/repository/bgruening/galaxy?tab=tags

PR #129, opened 2017-04-15
respect the GALAXY_VIRTUAL_ENV 

bgruening:
This will save us 200MB in all containers.

### DIFF
--- a/tasks/galaxy_root.yml
+++ b/tasks/galaxy_root.yml
@@ -1,10 +1,10 @@
 # TODO: following task is kind of a crutch - shouldn't be needed.
 - name: "Run common_startup.sh to create welcome.html from sample"
-  environment: 
-    GALAXY_VIRTUAL_ENV: "{{ galaxy_venv_dir }}"
   shell: scripts/common_startup.sh
-  arg: 
-    chdir: "{{ galaxy_venv_dir }}"
+  environment:
+    GALAXY_VIRTUAL_ENV: "{{ galaxy_venv_dir }}"
+  args:
+    chdir: "{{ galaxy_server_dir }}"
   become: True
   become_user: "{{ galaxy_user_name }}"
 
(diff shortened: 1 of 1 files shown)


mvdbeek:
Nice!

tasks/galaxy_root.yml
```diff
@@ -1,6 +1,6 @@
 # TODO: following task is kind of a crutch - shouldn't be needed.
 - name: "Run common_startup.sh to create welcome.html from sample"
-  shell: "cd {{ galaxy_server_dir }} && scripts/common_startup.sh"
+  shell: "export GALAXY_VIRTUAL_ENV={{ galaxy_venv_dir }} && cd {{ galaxy_server_dir }} && scripts/common_startup.sh"
```
mvdbeek: I think you could also write
```
environment: 
  GALAXY_VIRTUAL_ENV: {{ galaxy_venv_dir }}
```

mvdbeek: you can also drop the cd {{ galaxy_server_dir }} and
```
shell: scripts/common_startup.sh
args: 
  chdir: "{{ galaxy_venv_dir }}"
```

bgruening: I tried this before, but obviously I had a syntax error. It seems to work now. Thanks this is way better and more ansible like! :)

tasks/galaxy_root.yml
```diff
@@ -1,6 +1,10 @@
 # TODO: following task is kind of a crutch - shouldn't be needed.
 - name: "Run common_startup.sh to create welcome.html from sample"
-  shell: "cd {{ galaxy_server_dir }} && scripts/common_startup.sh"
+  environment: 
+    GALAXY_VIRTUAL_ENV: "{{ galaxy_venv_dir }}"
+  shell: scripts/common_startup.sh
+  arg:
```
mvdbeek: you're missing a `s` in `args`

bgruening:
ready to go!

jmchilton:
Very nice - thanks a bunch!

PR #132, opened 2017-05-02
Work on setting up Kubernetes and Condor job running environments.

jmchilton:
Fill in some other goodies (scripts and stuff) from docker-galaxy-stable.


templates/install_biojs_vis.sh.j2
```diff
@@ -0,0 +1,13 @@
+#!/bin/bash
+
```
bgruening: This one is experimental not sure if we should put it here.

PR #135, opened 2017-05-03
Improvements for Containerized Execution of Jobs

jmchilton:
[Updated to describe two commits.]

First commit enables using the ``docker_dispatch`` dynamic rule. GALAXY_DESTINATIONS_DEFAULT is the environment variable that sets the default job destination - this can be set "docker_dispatch" and then two new environment variables will be used - GALAXY_DESTINATIONS_DOCKER_DEFAULT and GALAXY_DESTINATIONS_NO_DOCKER_DEFAULT to determine the destinations for Docker and non-Docker jobs respectively.

The second commit allows fine grain determination of the container resolvers file used by Galaxy. I use this to enable direct annotation and existing BioContainers on quay - but disabling building containers on demand and checking the local Docker host for cached containers (since this isn't available in the Kubernetes setup).

### DIFF
--- a/templates/job_conf.xml.j2
+++ b/templates/job_conf.xml.j2
@@ -23,6 +23,7 @@
             <param id="k8s_persistent_volume_claim_mount_path">/export</param>
         </plugin>
 {% endif %}
+        <plugin id="local" type="runner" load="galaxy.jobs.runners.local:LocalJobRunner"/>
     </plugins>
     <!-- The default handler can be changed by specifying the GALAXY_HANDLERS_DEFAULT environment variable. -->
     <handlers default_from_environ="GALAXY_HANDLERS_DEFAULT" default="handlers">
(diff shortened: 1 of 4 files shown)


templates/job_conf.xml.j2
```diff
@@ -36,6 +37,16 @@
     </handlers>
     <!-- The default destination can be changed by specifying the GALAXY_DESTINATIONS_DEFAULT environment variable. -->
     <destinations default_from_environ="GALAXY_DESTINATIONS_DEFAULT" default="{{ galaxy_extras_galaxy_destination_default }}">
+        <destination id="docker_dispatch" runner="dynamic">
+            <!-- Allow different default destinations based on whether the tool
+                 supports Docker or not. -->
+            <param id="type">docker_dispatch</param>
+            <param id="docker_destination_id" from_environ="GALAXY_DESTINATIONS_DOCKER_DEFAULT">{{ galaxy_extras_galaxy_destination_docker_default }}</param>
```
bgruening: This is possible? Nice!

jmchilton: Should be 
 https://github.com/galaxyproject/galaxy/commit/0d457d30c105d78d120e2a8259cdc1264faf3833 - seems to work in my initial testing.

jmchilton: Yes - it totally works (well after the fixes in https://github.com/galaxyproject/galaxy/pull/4033)! I have a Kubernetes setup that is dispatching BioContainer jobs to Kubernetes "Job" API if the containers are available on quay.io and is executing the other tools locally with this these changes. You may want to consider a similar setup for Condor - run things in Docker when it makes sense and in a cluster mode if a container is unavailable.

jmchilton:
Taking this out of WIP - @bgruening any objections? Things are mostly off by default but I does add a local job runner that downstream projects can use - is that a problem?

templates/job_conf.xml.j2
bgruening: I was actually more thinking of running in a default container if no container is specified. Most of the tools without requirements are probably from Galaxy internal tools - so something like the https://quay.io/repository/biocontainers/coreutils?tab=tags should be enough or a standard Ubuntu image.

bgruening:
No objections, indeed I need this now to get Condor working!

templates/job_conf.xml.j2
jmchilton: A ton of tools require Galaxy though... I guess we do mount it?

jmchilton:
Thanks for the merge @bgruening !

templates/job_conf.xml.j2
bgruening: Yes. We mount Galaxy and if this does not help we could source the .venv.
I'm completely unsure what is the best way, my initial thought is just that an admins expects to run all tools in Docker and these exceptions are kind of in-transparent. But as you said the upload tool is really special :)

Issue #136, opened 2017-05-06
munge installation when ubuntu 16.04


drosofff:
As far as I can test the task https://github.com/galaxyproject/ansible-galaxy-extras/blob/master/tasks/slurm.yml#L7 fails when ubuntu 16.04.
The issue I see comes from failure in `apt-get install munge` and is related to https://bugs.launchpad.net/ubuntu/+source/munge/+bug/1287624/comments/11.

If before running the playbook I install manually munge (with failure) and fix it as indicated
```
1. sudo systemctl edit --system --full munge.service
   While in the editor, append " --syslog" to the ExecStart line:
   ExecStart=/usr/sbin/munged --syslog

2. sudo systemctl enable munge.service

3. sudo systemctl start munge.service
```
Then, the task  https://github.com/galaxyproject/ansible-galaxy-extras/blob/master/tasks/slurm.yml#L7 can run to completion.

Let me know if I missed something. If not I can work on a fix in the playbook

drosofff:
@bgruening @jmchilton Should I close the issue or is this procedure in your attributions ?

bgruening:
Feel free to close it :)

bgruening:
And thanks!

drosofff:
Happy to help !

PR #137, opened 2017-05-06
Fix installation of munge package in ubuntu 16.04

drosofff:
by pre-uploading a munge.service config file with
```
ExecStart=/usr/sbin/munged --syslog
```
The issue is detailed in #136

### DIFF
--- a/tasks/slurm.yml
+++ b/tasks/slurm.yml
@@ -4,10 +4,21 @@
     - python-pip
   when: galaxy_extras_install_packages
 
+- name: Create munge.service configuration file
+  template: src=munge.service.j2 dest=/etc/systemd/system/munge.service
+  when: galaxy_extras_install_packages
+
+- name: Install munge system package
+  apt: pkg=munge state={{ galaxy_extras_apt_package_state }}
+  when: galaxy_extras_install_packages
+
+- name: Restart munge service
+  systemd: name=munge daemon_reload=yes enabled=True state=started
+  when: galaxy_extras_install_packages
+
 - name: Install SLURM system packages
   apt: pkg={{ item }} state={{ galaxy_extras_apt_package_state }}
   with_items:
-    - munge
     - slurm-llnl
     - slurm-drmaa-dev
     - python-psutil
--- a/tasks/slurm.yml
+++ b/tasks/slurm.yml
@@ -6,7 +6,7 @@
 
 - name: Create munge.service configuration file
   template: src=munge.service.j2 dest=/etc/systemd/system/munge.service
-  when: galaxy_extras_install_packages
+  when: galaxy_extras_install_packages|bool and ansible_distribution_version >= "15.04"
 
 - name: Install munge system package
   apt: pkg=munge state={{ galaxy_extras_apt_package_state }}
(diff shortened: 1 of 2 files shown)


jmchilton:
Doesn't supervisor run munge though - or do you have that off in your settings?

drosofff:
Supervisor run munge is On in my settings.
However, whether or not supervisor run munge, the munge package is installed by apt before in the playbook and this does not work in 16.04. it also crashes the slurm packages installation by the way (still in 16.04)

jmchilton:
Okay - it seems to work in the Docker file planemo-machine but there we pre-install munge and slurm together. It also is working in the planemo machine VM which is 16.04 - so I don't understand - but I'll merge this and try another planemo machine build to see if it still works.

drosofff:
Thanks !

>  It also is working in the planemo machine VM which is 16.04 - so I don't understand

Maybe the 16.04 planemo machine uses a layer with pre-built munge/slurm ?

jmchilton:
So this does break planemo-machine https://github.com/galaxyproject/planemo-machine/pull/102.

```
TASK [galaxyprojectdotorg.galaxy-extras : Install munge system package] ********
ok: [localhost]
TASK [galaxyprojectdotorg.galaxy-extras : Restart munge service] ***************
fatal: [localhost]: FAILED! => {"changed": false, "failed": true, "msg": "failure 1 during daemon-reload: Failed to connect to bus: No such file or directory\n"}
```

Does this traceback to a particular ansible-galaxy-extras commit - was this working in the past for GalaxyKickStart?

drosofff:
Mmm. Sorry about that. I can't tell whether it used to work for GalaxyKickStart, I started testing it extensively in 16.04 only few days ago. But it's true that I remember a quick shot 2 months ago that failed but not because of munge.
Look, I think that my code could still work without restarting services. I'll let you know asap.

On another note, I see in https://travis-ci.org/galaxyproject/planemo-machine/jobs/229988623#L491 that munge and slurm are not installed in the same order as in https://github.com/galaxyproject/ansible-galaxy-extras/blob/master/tasks/slurm.yml. Maybe this matters ?

mvdbeek:
> On another note, I see in https://travis-ci.org/galaxyproject/planemo-machine/jobs/229988623#L491 that munge and slurm are not installed in the same order as in https://github.com/galaxyproject/ansible-galaxy-extras/blob/master/tasks/slurm.yml. Maybe this matters ?

That is ansible-galaxy-os, which we are using as well. We should probably fix it there, I already have https://github.com/galaxyproject/ansible-galaxy-os/pull/21 open, which still needs some work for debian 8. I'm wondering if it makes sense to include slurm in ansible-galaxy-os ... maybe we should make that at least conditional. I guess it makes more sense to install the package in ansible-galaxy-extras.

drosofff:
@jmchilton in the meantime I tested the modified slurm.yml without restarting service (which was not required indeed, adding a munge.service file before apt install munge is sufficient). Let me know if the new PR passes the tests in the planemo VM

dustin-cram:
I had this problem a while back and forgot to submit a fix - the issue is that Ubuntu insists on starting services the moment that they're apt installed, so it fails at that time.

My fix was to do a manual systemd mask (creating a symlink to /dev/null at /etc/systemd/system/munge.service) before installing munge.

I'm using GalaxyKickStart which has munge managed by supervisor, so in this case it can just remain masked.  Otherwise @drosofff's fix in https://github.com/galaxyproject/ansible-galaxy-extras/issues/136 can be applied and then unmasked.

jmchilton:
@drosofff That did indeed fix planemo-machine - thanks!

@dustin-cram Thanks for the additional information.

PR #138, opened 2017-05-07
use the galaxy wheel, this does not need a GCC installed locally

bgruening:


jmchilton:
There is a condor failure in the tests - it seems unrelated though right?

bgruening:
I hope so. I have major problems with htcondors debian repos. They always timeout since 2 days :(

PR #143, opened 2017-05-12
Change NGINX_X_ACCEL vars when using PROXY_PREFIX

abretaud:
Adapt 2 more env vars when the galaxy docker is running behind a proxy
Discussed in https://github.com/bgruening/docker-galaxy-stable/pull/310


bgruening:
Thanks @abretaud!

PR #144, opened 2017-05-14
introduce envsubst as a new way to change variables during startup

bgruening:
Some docs are included I would like to know if this is ok for this project as it makes things certainly more complicated. @erasche @mvdbeek @jmchilton 

@abretaud what do you think about this. It whould make proxy_prefix fully working, I think.

### DIFF
--- a/README.md
+++ b/README.md
@@ -12,8 +12,7 @@ Dependencies
 ------------
 
 This role assumes Galaxy has already been installed and configured
-(for instance with the [Galaxy
-role](https://github.com/galaxyproject/ansible-galaxy)).
+(for instance with the [Galaxy role](https://github.com/galaxyproject/ansible-galaxy)).
 
 Role variables
 --------------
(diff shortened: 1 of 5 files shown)


abretaud:
I can't test right now, but it looks great! 👍

mvdbeek:
I don't see why we're not using ansible, but if it's restricted to startup.sh.j2 it's fine.
For the nginx proxy prefix in the Docker variant of GalaxyKickStart we simply run the nginx tagged tasks on startup. Admittedly you'll need to have ansible installed, so if that's the reason I'd say why not.

mvdbeek:
Btw, we could add this https://github.com/arut/nginx-python-module to our nginx PPA, that should allow us to pull in env vars.

bgruening:
@mvdbeek thanks for looking into this. I'm not sure either if this is a good idea, but it is very light-weight and I don't need to carry around ansible, which is pretty heavy if you consider the entire python runtime env and additional libraries. The proposed concept was not for nginx alone and can be used by a lot more if people need to.

PR #147, opened 2017-05-28
Allow updating an existing user

hexylena:
The way this script is currently used in the docker image, it *should* update the existing user with a new password/api key. However, it doesn't.

This patch causes the script to update users even if they're already in the database. Thus, setting GALAXY_DEFAULT_ADMIN_PASSWORD and GALAXY_DEFAULT_ADMIN_KEY to different values at runtime will now work properly instead of silently failing to change the password/api key.

cc @bgruening

### DIFF
--- a/templates/create_galaxy_user.py.j2
+++ b/templates/create_galaxy_user.py.j2
@@ -34,6 +34,9 @@ def add_user(sa_session, security_agent, email, password, key=None, username="ad
             security_agent.user_set_default_permissions( user, history=True, dataset=True )
 
     if key is not None:
+        query = sa_session.query( APIKeys ).filter_by( user_id=user.id ).delete()
+        sa_session.flush()
+        
         api_key = APIKeys()
         api_key.user_id = user.id
         api_key.key = key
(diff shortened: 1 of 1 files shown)


erasche:
Don't merge yet, this has a bug I didn't catch earlier.

```
28.5.2017 16:45:07sqlalchemy.exc.IntegrityError: (psycopg2.IntegrityError) duplicate key value violates unique constraint "ix_api_keys_key"
28.5.2017 16:45:07DETAIL:  Key (key)=(admin) already exists.
28.5.2017 16:45:07 [SQL: 'INSERT INTO api_keys (create_time, user_id, key) VALUES (%(create_time)s, %(user_id)s, %(key)s) RETURNING api_keys.id'] [parameters: {'create_time': datetime.datetime(2017, 5, 28, 16, 45, 7, 460604), 'user_id': 1, 'key': 'admin'}]
28.5.2017 16:45:08==> /home/galaxy/logs/handler0.log <==
```

bgruening:
ping @erasche
Is this needed for GCC?

erasche:
no, we're using our own copy of this file. No need to merge immediately.

PR #149, opened 2017-06-07
Allow environment variables to conditionally enabled Galaxy job runners.

jmchilton:
These break non 17.09+ Galaxies so I had to introduce a minimum target Galaxy version for the role that is being executed. I defaulted this to 17.01 for now - anyone using an older Galaxy with master of this role?

### DIFF
--- a/templates/job_conf.xml.j2
+++ b/templates/job_conf.xml.j2
@@ -5,14 +5,24 @@
 {% if galaxy_extras_config_slurm %}
         <plugin id="slurm" type="runner" load="galaxy.jobs.runners.slurm:SlurmJobRunner">
             <param id="drmaa_library_path">/usr/lib/slurm-drmaa/lib/libdrmaa.so</param>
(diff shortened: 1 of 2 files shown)


templates/job_conf.xml.j2
```diff
@@ -72,6 +88,7 @@
   {% endif %}
   {% if galaxy_extras_config_k8_jobs %}
         <destination id="k8_default" runner="k8">
+            <param id="enabled" from_environ="GALAXY_RUNNERS_ENABLE_K8">true</param>
```
bgruening: Have you forgotten the galaxy_minimum_version here?

jmchilton: It doesn't break anything on the destinations - only the runners break if you pass them unknown parameters I think.

bgruening: Ok, fair enough!

PR #150, opened 2017-06-07
Allow many more job runner options at runtime.

jmchilton:
Flush out the job configuration file with many more job destinations, allowing different runtime defaults for the built containers using just the configuration variable GALAXY_DESTINATIONS_DEFAULT.

The options are:

- `local_no_container`: Use the local runner, do not use Docker.
- `local_docker`: Use the local runner along with Docker if a container for the tool can be found.
- `local_force_docker`: Use the local runner and force tools to use Docker*.
- `pbs_cluster`: Use PBS cluster without Docker.
- `pbs_cluster_docker`: Use PBS cluster with Docker if a container for the tool can be found.
- `pbs_cluster_force_docker`: Use PBS cluster and force tools to use Docker*.
- `slurm_cluster`: Use SLURM, do not use Docker.
- `slurm_cluster_docker`: Use SLURM and use Docker if a container for the tool can be found.
- `slurm_cluster_force_docker`: Use SLURM and force tools to use Docker*
- `condor_cluster`: Use Condor, do not use Docker.
- `condor_cluster_docker`: Use Condor, use Docker as part of job script if a container can be found.
- `condor_cluster_force_docker`: Use Condor, force Docker usage*.
- `condor_docker_universe`: Use Condor through Docker universe*.
- `condor_docker_universe_dispatch`: Use Condor through Docker universe if a container for the tool can be found, else use the vanilla universe.
- `k8_default`: Run through Docker on Kubernetes*.
- `k8_or_local_dispatch`: If a container for the tool can be found, run on Kubernetes else on local runner (`local_no_container`).
- `k8_or_slurm_dispatch`: If a container for the tool can be found, run on Kubernetes else on SLURM (`slurm_cluster`).
- `k8_or_condor_dispatch`: If a container for the tool can be found, run on Kubernetes else on Condor as a cluster job (`condor_cluster`).
- `docker_dispatch`: Dispatch between any two using the enironment variables `GALAXY_DESTINATIONS_DOCKER_DEFAULT` and `GALAXY_DESTINATIONS_NO_DOCKER_DEFAULT`.

\* `GALAXY_DOCKER_DEFAULT_CONTAINER` will be used for the default container if no tool-specific container is located.

Subtlety breaks backward compatibility for certain edge case configurations - probably only affecting docker-galaxy-stable?.

### DIFF
--- a/templates/job_conf.xml.j2
+++ b/templates/job_conf.xml.j2
@@ -57,13 +57,13 @@
             <param id="docker_destination_id" from_environ="GALAXY_DESTINATIONS_DOCKER_DEFAULT">{{ galaxy_extras_galaxy_destination_docker_default }}</param>
             <param id="default_destination_id" from_environ="GALAXY_DESTINATIONS_NO_DOCKER_DEFAULT">{{ galaxy_extras_galaxy_destination_default }}</param>
         </destination>
-        {{ macros.destination("local_no_container", "local") }}
-        {{ macros.destination("local_docker", "local", docker=True) }}
-        {{ macros.destination("local_force_docker", "local", docker=True, force_docker=True) }}
+        {% call macros.destination("local_no_container", "local") %}{% endcall %}
+        {% call macros.destination("local_docker", "local", docker=True) %}{% endcall %}
+        {% call macros.destination("local_force_docker", "local", docker=True, force_docker=True) %}{% endcall %}
 {% if galaxy_extras_config_pbs %}
-        {{ macros.destination("pbs_cluster", "pbs") }}
-        {{ macros.destination("pbs_cluster_docker", "pbs", docker=True) }}
-        {{ macros.destination("pbs_cluster_force_docker", "pbs", docker=True, force_docker=True) }}
+        {% call macros.destination("pbs_cluster", "pbs") %}{% endcall %}
+        {% call macros.destination("pbs_cluster_docker", "pbs", docker=True) %}{% endcall %}
+        {% call macros.destination("pbs_cluster_force_docker", "pbs", docker=True, force_docker=True) %}{% endcall %}
 {% endif %}
 {% if galaxy_extras_config_slurm %}
         {% call macros.destination("slurm_cluster", "slurm") %}
@@ -98,7 +98,7 @@
         {{ macros.docker_dispatch_destination("condor_docker_universe_dispatch", "condor_docker_universe", "condor_cluster")}}
 {% endif %}
 {% if galaxy_extras_config_k8_jobs %}
-        {{ macros.destination("k8_default", "k8", docker=True, force_docker=True) }}
+        {% call macros.destination("k8_default", "k8", docker=True, force_docker=True) %}{% endcall %}
         {{ macros.docker_dispatch_destination("k8_or_local_dispatch", "k8_default", "local_no_container")}}
         {{ macros.docker_dispatch_destination("k8_or_slurm_dispatch", "k8_default", "slurm_cluster")}}
         {{ macros.docker_dispatch_destination("k8_or_condor_dispatch", "k8_default", "condor_cluster")}}
--- a/templates/job_conf.xml.j2
+++ b/templates/job_conf.xml.j2
@@ -60,30 +60,107 @@
         <destination id="local_no_container" runner="local">
             <env file="{{ galaxy_venv_dir }}/bin/activate"/>
         </destination>
+        <destination id="local_docker" runner="local">
+            <env file="{{ galaxy_venv_dir }}/bin/activate"/>
+            {{ macros.docker_enabled() }}
+        </destination>
+        <destination id="local_force_docker" runner="local">
+            <env file="{{ galaxy_venv_dir }}/bin/activate"/>
+            {{ macros.docker_enabled(True) }}
+        </destination>
+        <destination id="local_docker_dispatch" runner="dynamic">
+            <param id="type">docker_dispatch</param>
+            <param id="docker_destination_id">local_docker</param>
+            <param id="default_destination_id">local_no_docker</param>
+        </destination>
 {% if galaxy_extras_config_slurm or galaxy_extras_config_pbs or galaxy_extras_config_condor %}
   {% if galaxy_extras_config_pbs %}
         <destination id="pbs_cluster" runner="pbs">
             <param id="enabled" from_environ="GALAXY_RUNNERS_ENABLE_PBS">true</param>
             <env file="{{ galaxy_venv_dir }}/bin/activate"/>
+        </destination>
+        <destination id="pbs_cluster_docker" runner="pbs">
+            <param id="enabled" from_environ="GALAXY_RUNNERS_ENABLE_PBS">true</param>
+            <env file="{{ galaxy_venv_dir }}/bin/activate"/>
             {{ macros.docker_enabled() }}
         </destination>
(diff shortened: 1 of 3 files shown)


templates/job_conf.xml.j2
```diff
@@ -60,30 +60,107 @@
...
+            {{ macros.docker_enabled() }}
+        </destination>
+        <destination id="local_force_docker" runner="local">
+            <env file="{{ galaxy_venv_dir }}/bin/activate"/>
+            {{ macros.docker_enabled(True) }}
+        </destination>
+        <destination id="local_docker_dispatch" runner="dynamic">
+            <param id="type">docker_dispatch</param>
+            <param id="docker_destination_id">local_docker</param>
+            <param id="default_destination_id">local_no_docker</param>
```
bgruening: Should this be `local_no_container`? Below as well.

templates/job_conf.xml.j2
```diff
@@ -60,30 +60,107 @@
         <destination id="local_no_container" runner="local">
             <env file="{{ galaxy_venv_dir }}/bin/activate"/>
         </destination>
+        <destination id="local_docker" runner="local">
+            <env file="{{ galaxy_venv_dir }}/bin/activate"/>
+            {{ macros.docker_enabled() }}
+        </destination>
+        <destination id="local_force_docker" runner="local">
+            <env file="{{ galaxy_venv_dir }}/bin/activate"/>
```
bgruening: Should we put this into a macro as well? `<env file="{{ galaxy_venv_dir }}/bin/activate"/>`

templates/job_conf.xml.j2
```diff
@@ -60,30 +60,107 @@
...
+        <destination id="condor_cluster_force_docker" runner="condor">
+            <env file="{{ galaxy_venv_dir }}/bin/activate"/>
             <param id="universe" from_environ="GALAXY_CONDOR_UNIVERSE">vanilla</param>
-        {% endif %}
+            <param id="enabled" from_environ="GALAXY_RUNNERS_ENABLE_CONDOR">true</param>
+            {{ macros.docker_enabled(True) }}
+        </destination>
+        <destination id="condor_docker_universe" runner="condor">
             <env file="{{ galaxy_venv_dir }}/bin/activate"/>
+            <param id="universe" from_environ="GALAXY_CONDOR_UNIVERSE">docker</param>
```
bgruening: Wondering if the `<env>` here is missing and if we can skip this entire section as we can change the universe via runtime already. Also see my general notes about this setup.

templates/job_conf.xml.j2
```diff
@@ -60,30 +60,107 @@
...
+        -->
+        <destination id="condor_docker_cluster_dispatch" runner="dynamic">
+            <param id="type">docker_dispatch</param>
+            <param id="docker_destination_id">condor_cluster_docker</param>
+            <param id="default_destination_id">condor_cluster</param>
+        </destination>
+        <destination id="condor_docker_universe_dispatch" runner="dynamic">
+            <param id="type">docker_dispatch</param>
+            <param id="docker_destination_id">condor_docker_universe</param>
+            <param id="default_destination_id">condor_cluster</param>
```
bgruening: Continuing from above if I set `GALAXY_CONDOR_UNIVERSE="docker"` and choose `condor_docker_cluster_dispatch` I get the same as `condor_docker_universe_dispatch` isn't it?

templates/job_conf.xml.j2
```diff
@@ -92,6 +169,22 @@
             <!-- Following parameter must be set to True for this runner. -->
             <param id="docker_enabled">true</param>
             <env file="{{ galaxy_venv_dir }}/bin/activate"/>
+            <param id="docker_default_container_id" from_environ="GALAXY_DOCKER_DEFAULT_CONTAINER">{{ galaxy_docker_default_image }}</param>
+        </destination>
+        <destination id="k8_or_local_dispatch" runner="dynamic">
+            <param id="type">docker_dispatch</param>
+            <param id="docker_destination_id">k8_default</param>
+            <param id="default_destination_id">local_no_docker</param>
```
bgruening: local no container?

bgruening:
:100: 

* I would remove `condor_docker_universe` I tried a few days to get it work properly, but there seems to be many unresolved questions on the ht-condor side. For example you as admin need to define volumes in advance and so on. If we don't want to make this list comprehensive I would rather not put it in.

* `k8_or_local_dispatch` is there a real world use-case for this?

jmchilton:
- > Should we put this into a macro as well? `<env file="{{ galaxy_venv_dir }}/bin/activate"/>`

  I guess? I've redone everything to be much more macro heavy - it is more dry and less direct now.
- > Should this be local_no_container? Below as well.

  Fix this.
- I'm disappointed about `condor_docker_universe` - I thought that was being used. Does it not work in the context of docker-galaxy-stable or not at all really? I'd rather keep it and encourage people to hack on it - maybe just document it as beta - but I can drop it if you'd prefer.
- `k8_or_local_dispatch` - It is a good question if this has a real world use case. I used the local runner in testing because it had fewer moving parts and stuff - and for small non-docker jobs like upload and history export it made sense to me to use it. In the docker-galaxy-stable setup you have good testing of SLURM for instance and so it might always be preferable to do that. Maybe in other setups this isn't the case? I feel like this is a fine target I guess - though in most production settings it would be better to use `k8_or_slurm` I suspect.
- > if we can skip this entire section as we can change the universe via runtime already

  All the K8 switches can be done with the existing single dispatch and setting up two environment also - I just sort of think it is easier to convey in a big table having a single environment variable?

bgruening:
> I'm disappointed about condor_docker_universe - I thought that was being used. Does it not work in the context of docker-galaxy-stable or not at all really? I'd rather keep it and encourage people to hack on it - maybe just document it as beta - but I can drop it if you'd prefer.

The vanilla runner is in use afaik. That the docker universe is not that useful in Galaxy context is not really Galaxy fault, it is just very hard to set up. I was able to get it running but than you have to declare mount points that more or less mount the entire database directory into a container ... so not really useful.

Thanks for fixing!

templates/macros.xml.j2
```diff
@@ -1,11 +1,27 @@
-{% macro docker_enabled() -%}
+{% macro destination(id, runner, docker=False, force_docker=False) -%}
+    <destination id="{{ id }}" runner="{{ runner }}">
+        <env file="{{ galaxy_venv_dir }}/bin/activate"/>
     {% if galaxy_source_shellrc %}
         <env file="{{ galaxy_user_shellrc }}" />
     {% endif %}
-        <param id="docker_enabled" from_environ="GALAXY_DOCKER_ENABLED">{{ galaxy_docker_enabled }}</param>
+        <param id="enabled" from_environ="GALAXY_RUNNERS_ENABLE_{{ runner|upper }}">true</param>
+        <param id="docker_enabled">true</param>
```
bgruening:
```
{% if docker %}
    <param id="docker_enabled">true</param>
{% endif %}
```

???

jmchilton: Yes - that is what I meant - thanks!

bgruening:
@jmchilton is this one still WIP?
@mvdbeek any objection to merge the PR as is?

bgruening:
This is now integrated into Galaxy Docker and the the Singularity part is under testing.

https://github.com/bgruening/docker-galaxy-stable/pull/349
https://travis-ci.org/bgruening/docker-galaxy-stable/jobs/242178166

PR #152, opened 2017-06-08
Increase buffer size for uWSGI

manabuishii:
Default buffer size in uWSGI is 4096.
But it's small for long error message.

In my case , I try to import workflow  to clean docker galaxy instance ,
the galaxy instance has no tool for the workflow.
Error message is long , like this

```
/admin?status=error&message=Imported%2C+but+some+steps+in+this+workflow+have+validation+err
```
uwsgi.log error message has like this

```
invalid request block size: 4098 (max 4096).
```

but increase buffer size , there is no uwsgi error

### DIFF
--- a/templates/nginx.conf.j2
+++ b/templates/nginx.conf.j2
@@ -9,6 +9,10 @@ events {
 http {
     include       mime.types;
     default_type  application/octet-stream;
+    
+    uwsgi_buffers 8 16384;
+    uwsgi_buffer_size 16384;
+
 
     sendfile        on;
 
--- a/templates/supervisor.conf.j2
+++ b/templates/supervisor.conf.j2
@@ -102,9 +102,9 @@ priority        = 200
 [program:galaxy_web]
 {% if galaxy_uwsgi|bool %}
 {% if galaxy_uwsgi_static_conf|bool %}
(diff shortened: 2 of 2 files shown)


bgruening:
@manabuishii thanks for the patch. I'm wondering why usegalaxy.org is not having something similar in its configuration. @natefoo could you please have a look?

manabuishii:
I use bgruening/galaxy-stable:dev pulled on 2017-06-07

jmchilton:
@bgruening I think @natefoo did do this https://github.com/galaxyproject/galaxy/issues/1530.

bgruening:
Thanks John! I found it. @manabuishii can you use the same value as Nate?
https://github.com/galaxyproject/usegalaxy-playbook/blob/dc487e2a2f585554dc5140660c2536e505eedca9/templates/galaxy/usegalaxy.org/config/uwsgi.ini.j2#L11

manabuishii:
@bgruening @jmchilton Thanks !

I look at your comments.

The problem is happened on Docker version Galaxy .
And to resolve it, 
not only supervisord option but also nginx settings need more configuration.

I test 16384 (the same value https://github.com/galaxyproject/usegalaxy-playbook/blob/dc487e2a2f585554dc5140660c2536e505eedca9/templates/galaxy/usegalaxy.org/config/uwsgi.ini.j2#L11 )


Supervisor option resolve nginx -> uwsgi(galaxy)

```
-b 16384
```

But there is still problem uwsgi(galaxy) -> nginx

needs more option to nginx.conf http section

```
    uwsgi_buffers 8 16384;
    uwsgi_buffer_size 16384;
```

These two configuration update has no problem when I import workflow.

I will update this pull request or create another pull request.

bgruening:
Thanks @manabuishii!

PR #154, opened 2017-06-11
Add initial singularity support

bgruening:
@jmchilton this add singularity to SLRUM, we can add many more later.

I have tried to unify the macros even more, but ended up with the current solution, it was getting to ugly with nested variables and so on. However, I ended up with `container_type` and `force_container` to make it more general.

### DIFF
--- a/defaults/main.yml
+++ b/defaults/main.yml
@@ -80,7 +80,7 @@ galaxy_singularity_sudo: false
 galaxy_singularity_default_image: ''
 galaxy_singularity_volumes_from: ""
 # rw directories are not considered if the parent is ro
-galaxy_singularity_volumes : "$tool_directory:ro,$working_directory:rw,$default_file_path:rw"
+galaxy_singularity_volumes : "$defauls"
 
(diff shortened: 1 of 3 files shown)


bgruening:
Please note that it is targeting your branch.

defaults/main.yml
```diff
@@ -79,7 +79,8 @@ galaxy_singularity_sudo: false
 # ToDo create default image
 galaxy_singularity_default_image: ''
 galaxy_singularity_volumes_from: ""
-galaxy_singularity_volumes : "$defaults"
+# rw directories are not considered if the parent is ro
+galaxy_singularity_volumes : "$tool_directory:ro,$working_directory:rw,$default_file_path:rw"
```
jmchilton: xref https://github.com/galaxyproject/galaxy/pull/4180 - does that fix the need for this?

bgruening:
@jmchilton ok to merge?

jmchilton:
Truly awesome - thanks so much @bgruening!

PR #155, opened 2017-06-18
Add 500 error and modify error pages location

angellandros:


defaults/main.yml
```diff
@@ -153,7 +153,7 @@ nginx_welcome_location: "{{ nginx_prefix_location }}/etc/galaxy/web"
 nginx_welcome_path: "/etc/galaxy/web"
 
 # Synchronize error handling with ansible-galaxy role.
-galaxy_errordocs_dest: "/root/"
+galaxy_errordocs_dest: "/root/errordocs"
```
bgruening: Why is this change needed?

templates/nginx.conf.j2
```diff
@@ -199,6 +199,7 @@ http {
         # error docs
         error_page  502 503 504 {{ nginx_prefix_location }}/error/502/index.shtml;
         error_page  413         {{ nginx_prefix_location }}/error/413/index.html;
+	    error_page  500         {{ nginx_prefix_location }}/error/500/index.html;
```
bgruening: Indentation is wrong here.

defaults/main.yml
angellandros: In the NGINX config file, `{{ nginx_prefix_location }}/error` is mapped to `{{ galaxy_errordocs_dest }}` which is `/root/`. So, for example, in the case of error 502, NGINX tried to load `/root//502/index.shtml`. I have added `errors` to the end of the path, to eliminate the double slash, and also keep all the error pages in a separated folder, instead of in the root.

templates/nginx.conf.j2
angellandros: PyCharm's fault. Fixed it with VIM.

bgruening:
Looks, good. Travis has some problems today. Let's try to trigger a test.

bgruening:
Thanks @angellandros!

jmchilton:
This won't mask the proxies 500 error page will it? Frequently with Galaxy those are helpful.

PR #157, opened 2017-06-26
Extract and isolate @afgane's CVMFS feature for galaxyextras

chambm:
* simplify repository list to just data.galaxyproject.org
* add `galaxy_extras_config_cvmfs` option, defaults to false
* add supervisord entry for `autofs`, which CVMFS requires for plug'n'play operation

The autofs stuff should probably be conditional on CVMFS. I wasn't sure whether keeping all 3 stratum1 servers was correct or not, since I removed the test.* and main.* stratum0 servers.


defaults/main.yml
```diff
@@ -219,6 +220,7 @@ supervisor_manage_reports: true
 supervisor_manage_ie_proxy: true
 supervisor_manage_condor: true
 supervisor_manage_cron: true
+supervisor_manage_autofs: true
```
bgruening: You need to install automount/autofs isn't it?

bgruening:
I think we fixed the tests. Restarting.

defaults/main.yml
chambm: I think autofs will come installed by default on Ubuntu 14.04; is the role supposed to be more portable than that?

bgruening: Yes. And it is not installed on mine 16.04 ... but maybe this is just my instance.

bgruening:
Here is something wired. Some commits are in this branch which are already in master :(
Can you try to merge master into your branch?

chambm:
I must have screwed up the rebase. If you squash on merge does that matter?

bgruening:
Not sure, I don't want to try this :)

bgruening:
@chambm this one is obsolete?

chambm:
Yes, replaced by #160.

PR #160, opened 2017-06-30
Extract and isolate @afgane's CVMFS feature for galaxyextras

chambm:
* simplify repository list to just data.galaxyproject.org
* add galaxy_extras_config_cvmfs option, defaults to false
* add supervisord entry for autofs, which CVMFS requires for plug'n'play operation

The autofs stuff should probably be conditional on CVMFS. I wasn't sure whether keeping all 3 stratum1 servers was correct or not, since I removed the test.* and main.* stratum0 servers.


tasks/cvmfs_client.yml
```diff
@@ -0,0 +1,73 @@
...
+  file: path=/galaxy state=directory
+
+# This is required to be able to use Galaxy .len files from Main w/o mods
+- name: Create a symlink from CVMFS to /galaxy/data
+  file: src=/cvmfs/data.galaxyproject.org/byhand dest=/galaxy/data state=link force=yes
+
+- name: Use tool_data_table_conf.xml from usegalaxy-playbook
+  get_url:
+    url: https://raw.githubusercontent.com/galaxyproject/usegalaxy-playbook/067433f547f67b848f2e4597274a379f06da04b2/files/galaxy/usegalaxy.org/config/tool_data_table_conf.xml
+    dest: "{{ galaxy_server_dir }}/config/tool_data_table_conf.xml"
```
bgruening: Can we use {{ galaxy_config_dir }} here?

chambm: I tried that first, and it doesn't work with your docker image because you change that to /export/etc/galaxy. So either I could add another symbolic link from config to there, or just use the config path directly; went with the latter approach.

chambm:
OK @bgruening where do I set `GALAXY_CONFIG_TOOL_DATA_TABLE_CONFIG_PATH=/etc/galaxy/tool_data_table.xml` ? In .travis.yml?

defaults/main.yml
```diff
@@ -22,6 +22,7 @@ galaxy_extras_config_ssl_method: self-signed  # This may be 'own', 'self-signed'
 galaxy_extras_galaxy_domain: "localhost"  # This is used by letsencrypt, set it to the domain name under which galaxy can be reached
 galaxy_extras_config_startup: true
 galaxy_extras_config_rabbitmq: false
+galaxy_extras_config_cvmfs: &galaxy_extras_config_cvmfs true
```
afgane: Is there a reason to use the anchor tag? The rest of the file uses interpolated variables so perhaps better to stay consistent...

chambm: Will interpolated variables work with boolean? Or just strings?

afgane: They seem to be working fine with other vars, e.g., https://github.com/galaxyproject/ansible-galaxy-extras/blob/master/defaults/main.yml#L34

chambm: OK no problem to change to interpolation. Do you know why your VM had automount already running, i.e. how can we detect that?

drosofff:
Hi @chambm & @afgane 
my quick and probably weak input on the issue with GalaxyKickStart

It true that the install finishes with the supervisor service autofs in FATAL state, and I have the same "/usr/sbin/automount: program is already running." warning in the log. Thus my guess is that ansible should stop the autofs service before to transfer its control to supervisorctl.

There is also probably another facet of the same issue (yet I am not sure it's the same issue) in the Travis test of GalaxyKickStart, for instance here where I tested this branch:
https://travis-ci.org/ARTbio/GalaxyKickStart/builds/261070992#L1456

Here the context is that GalaxyKickStart run on the the top of a Docker Container artbio/ansible-galaxy-os to speed up the test. From the log, I would say that there is a problem of rights or of access to config files for the installation of the autofs and nfs-common packages, but I did not think enough time to figure out precisely what is happening.

Hope it will provide you with some useful hints ! Cheers

chambm:
Er, what is going on here?

afgane:
I've just restarted the tests in hope this was a fluke.

bgruening:
What do you mean? Travis seems to be busy. The entire Galaxyproject has only X travis jobs to run in parallel :(

chambm:
So tests can start running but fail because there aren't enough free resources or something?

bgruening:
Oh, I have only seen the waiting travis and was thinking that it does not start fast enough :)
What was the error?

chambm:
Yeah @afgane hid the recent error by restarting the build. Let's see if it happens again.

afgane:
Every request to Galaxy was getting `Connection refused` response so something seems to have gone off rails and Galaxy didn't start. We'll see shortly.

chambm:
Great. So how to test this autofs issue? Or forget about testing it and just try to make the supervisor job tolerant to autofs already running? Then again, maybe having the supervisor job be FATAL if it's already running isn't such a big deal. As long as CVMFS is working, it seems like a minor detail.

afgane:
@chambm I think just removing the service file that comes with the package resolves the issue. I have a local commit available if you can just give me write access to this branch?

chambm:
You mean removing the supervisor j2 file? I'm pretty sure starting autofs is necessary in many plain docker cases. There must be a way to support both ways (i.e. autofs already started and autofs installed but not running).

bgruening:
@chambm I think was Enis is saying you can remove the service file for autofs, this is the file that is debian specific and starts a deamon. By removing it autofs simply do not start anymore.

`systemctl disable autofs` could work on systemd based systems.

bgruening:
@afgane, @chambm it's green :)

afgane:
I see only 👍 's on the screen → I guess that's a clear merge invite! ;)

bgruening:
@chambm thanks a lot, I will update the galaxy-stable branch and try to work there.
@afgane thanks for the merge and the help!

defaults/main.yml
```diff
@@ -252,3 +255,35 @@ nginx_ssl_certificate: /etc/ssl/private/nginx.crt
 # This defaults to a set of self-generated keys
 src_nginx_ssl_certificate_key: nginx_sample.key
 src_nginx_ssl_certificate: nginx_sample.crt
+
+cvmfs_deb_url: "https://ecsft.cern.ch/dist/cvmfs/cvmfs-2.1.20/cvmfs_2.1.20_amd64.deb"
```
natefoo: This should probably be 2.3.5.

defaults/main.yml
```diff
@@ -252,3 +255,35 @@ nginx_ssl_certificate: /etc/ssl/private/nginx.crt
...
+      GzfDoswr5TTk7aH/FIXUjLnLGGCOzPtUC244IhHARzu86bWYxQJUw0/kZl5wVGcH
+      maSgr39h1xPst0Vx1keJ95AH0wqxPbCcyBGtF1L6HQlLidmoIDqcCQpLsGJJEoOs
+      NVNhhcb66OJHah5ppI1N3cZehdaKyr1XcF9eedwLFTvuiwTn6qMmttT/tHX7rcxT
+      owIDAQAB
+      -----END PUBLIC KEY-----
+
+cvmfs_server_urls:
+  - domain: galaxyproject.org
+    urls:
+      - "http://cvmfs1-psu0.galaxyproject.org/cvmfs/@fqrn@"
```
natefoo: cvmfs1-iu0 and cvmfs-tacc0 (.galaxyproject.org) also serve this repo

defaults/main.yml
bgruening: Why not https://ecsft.cern.ch/dist/cvmfs/cvmfs-2.4.1/cvmfs_2.4.1~1+ubuntu14.04_amd64.deb ?

tasks/cvmfs_client.yml
```diff
@@ -0,0 +1,78 @@
...
+    mode: "0644"
+  with_items: "{{ cvmfs_server_urls }}"
+
+- name: Configure CernVM-FS client settings
+  copy:
+    content: |
+      CVMFS_REPOSITORIES="{%- for repo in cvmfs_repositories -%}{{ ',' if loop.index0 > 0 else '' }}{{ repo.repository }}{%- endfor -%}"
+      CVMFS_HTTP_PROXY="{{ cvmfs_http_proxies | join(';') }}"
+      CVMFS_QUOTA_LIMIT="{{ cvmfs_quota_limit | default('4000') }}"
+    dest: "/etc/cvmfs/default.local"
```
natefoo: It's probably safe to add `CVMFS_USE_GEOAPI={{ cvmfs_use_geoapi | default('yes') }}` as well.

natefoo:
Sorry for the post-merge review, I added a few minor comments.

defaults/main.yml
bgruening: 2.4.1 in general. We need to change the URL according to the used ubuntu version.

chambm: Or other operating system. I found myself sed'ing this file yesterday in order to re-use the role on an Amazon Linux system (rpm-based).

defaults/main.yml
chambm: That's been added in a later PR.

tasks/cvmfs_client.yml
chambm: Do you think geoapi or `cvms_talk host probe` is better?

defaults/main.yml
natefoo: I suggested 2.3.5 only because 2.4.1 hasn't been announced (but perhaps it is considered "released").

defaults/main.yml
natefoo: Whoops, thanks.

tasks/cvmfs_client.yml
natefoo: It depends, `cvmfs_talk host probe` will probably result in better performance, but GeoAPI is probably going to distribute the load a bit better.

Also the documentation is somewhat lacking on `cvmfs_talk host probe`, so I'm not sure if it actually does the reordering itself at runtime, and whether the ordering stays that way until the next time `cvmfs_talk host probe` is run (in which case you'd want to run it periodically from cron)?

defaults/main.yml
bgruening: Addressed in https://github.com/galaxyproject/ansible-galaxy-extras/pull/168

tasks/cvmfs_client.yml
bgruening: Addressed in https://github.com/galaxyproject/ansible-galaxy-extras/pull/168

PR #161, opened 2017-07-28
Enable loading of conditional requirements in /usr/bin/startup

rhpvorderman:
Plain galaxy starts using [common_startup.sh](https://github.com/galaxyproject/galaxy/blob/dev/scripts/common_startup.sh) which loads conditional requirements. For example if you have specified an auth_con.xml file in galaxy.ini it will load the python-ldap package. 
Unfortunately this feature was not present in docker-galaxy-stable. 
I copied the responsible script in common_startup.sh to startup.sh.j2 and tested it in the docker image by trying to authenticate on our own ldap server. 
It seems to work. All that required is that the FETCH_WHEELS environment variable is set to something other than "". If further modifications need to be made to properly integrate this please let me know.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -51,4 +51,6 @@ script:
   - make test_bioblend
   #- make test_docker_in_docker
   # Test the conditional loading of dependencies.
-  - bash ./tests/conditional_deps/test_script.sh
+  - echo $PWD
+  - ls
+  - bash tests/conditional_deps/test_script.sh
--- a/.travis.yml
+++ b/.travis.yml
@@ -51,16 +51,4 @@ script:
   - make test_bioblend
   #- make test_docker_in_docker
   # Test the conditional loading of dependencies.
-  - >
-    docker run -d -p 8080:80 -p 8021:21 -p 8022:22
-    --name galaxy_test_container
-    -e GALAXY_CONFIG_AUTH_CONFIG_FILE=config/auth_conf.xml
-    -v tests/conditional_deps/auth_conf.xml:/galaxy-central/config/auth_conf.xml
-    --name galaxy-test
-    galaxy-docker/test
-  - docker ps
-  - sleep 30
-  - >
-    docker exec -u 1450 galaxy-test /galaxy_venv/bin/pip list --format=columns | grep python -ldap
-    if [$? == 0 ]; then echo "Conditional dependency loaded";
-    else echo "Conditional dependency not loaded!" && exit 1 ; fi
+  - test/conditional_deps/test_scripts.sh
--- a/.travis.yml
+++ b/.travis.yml
@@ -51,4 +51,4 @@ script:
   - make test_bioblend
   #- make test_docker_in_docker
   # Test the conditional loading of dependencies.
-  - bash ./tests/conditional_deps/test_scripts.sh
+  - bash ./tests/conditional_deps/test_script.sh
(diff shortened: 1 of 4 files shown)


bgruening:
Stupid quesiton, but why is this not loaded by default?
Can you add a simple test for this and add the new variables to the docs?

Thanks!

rhpvorderman:
I added the test. 
Unfortunately the master branch fails to build the docker image, so all subsequent tests fail. Can this be fixed?

bgruening:
I'm not sure why this is failing so randomly, restarting often helps.

jmchilton:
Would it be better to just unconditionally install the optional dependencies like python-ldap prior into the container? I think I would prefer that solution - it seems more stable than doing at runtime.

rhpvorderman:
> Would it be better to just unconditionally install the optional dependencies like python-ldap prior into the container? I think I would prefer that solution - it seems more stable than doing at runtime.

@jmchilton Yes. It would. But unfortunately this does not work. There is for example a python package for the integration with a MySQL database. This will only install if the base system has MySQL installed. So using `pip install -r lib/galaxy/dependencies/conditional-requirements.txt` will fail. The workaround would be to only install packages without such depencies to catch the majority of use cases, but then we would be maintaining two separate lists and if a MySQL database is used, this will give an error anyway. So this solution does not capture all use cases.

Using the same script as common_startup.sh captures all use cases. And it is already used in galaxy main. So I guess stability is thoroughly tested already.

jmchilton:
@rhpvorderman I'd argue that if we wanted to support MySQL we should pre-install the dependencies for that wheel and then install that wheel at image build time. I don't think we want to support MySQL though and I don't think it would work to do it this way because we don't provide the hook needed to pre-install the apt packages needed.

Is something like that needed for python-ldap?

The way Galaxy does conditional requirements does cover a lot of cases - but the goal with it is to make easy and automatic - not to serve as best practices. They do a lot of things in ways they shouldn't operate in production environments.

- Ideally, applications like Galaxy should be able to work with the application itself and its dependencies mounted read-only - and the ideal production environment is that. I think Nate has worked pretty hard to mount usegalaxy.org's Galaxy this way and it is a good thing.
- Ideally, containers wouldn't need to install things at runtime - they should operate properly without an internet connection.
- Ideally the container should ship with all the dependencies it needs I think.

rhpvorderman:
@jmchilton You are right. It is better to include the dependencies in the docker container beforehand. I'd argue that it is still beneficial to keep this script change as a backup in case dependencies are upgraded on a new release of galaxy. Also when an external virtualenv is used with the docker container this can be useful.
I suggest that we keep this script and that I open another pull request for a role that populates the galaxy venv with most optional dependencies.

bgruening:
ping @jmchilton @mvdbeek for a small review.
Imho mostly ok with this, also I agree with John and I think if we are going to ship this by default in the container these deps should not be installed by default-startup. So can we change the NONUSE thing so that the deps are only installed when we explicitly request this?

rhpvorderman:
Abandoned the NONUSE script in favour of LOAD_GALAXY_CONDITIONAL_DEPENDENCIES.

bgruening:
@jmchilton is this ok for you?
It adds functionality and tests and does not conflict with other functionalities afaik.

jmchilton:
Yeah - I like the swap of the default. This looks really good to me now - thanks @rhpvorderman and sorry for the delay. Also thanks for the tests - that is awesome!

bgruening:
Thanks for the merge John! And thanks for the nice feature @rhpvorderman!

Issue #164, opened 2017-08-09
Install autofs system package breaks GalaxyKickStart


drosofff:
@rhpvorderman @afgane 
To let you know that the PR #160 broke the GalaxyKickStart Travis tests.
Stricto sensus the PR works, except that supervisor autofs is in FATAL states. However, the test fails because we build a docker container starting from [artbio/ansible-galaxy-os](https://hub.docker.com/r/artbio/ansible-galaxy-os/~/dockerfile/): here, the task `[galaxyprojectdotorg.galaxy-extras : Install autofs system package]`fails due to a failure in autofs package installation (see https://travis-ci.org/ARTbio/GalaxyKickStart/builds/261895456#L1443).

I'll try to work at a fix of this. In the meantime, I will pin ansible-galaxy-extras role to commit 03e894feface65f9b4c0df1a41fb70c1fe8db159. What would be cool in addition would be to condition cvmfs_client.yml execution with another variable than galaxy_extras_install_packages. That way, GKS could easily skip the cvmfs stuff that is not currently required.

afgane:
There is a conditional for setting up CVMFS in `defaults`: https://github.com/galaxyproject/ansible-galaxy-extras/blob/master/defaults/main.yml#L25

If you build the container by running the tests locally, can you access it after the process fails? The error log is not very helpful as to what's failing (it just says `"* Starting automount...", "   ...fail!"`).

I guess I'll also have to revisit the `FATAL` state because I thought I had fixed it in the last couple of runs I did...

drosofff:
> There is a conditional for setting up CVMFS in defaults: https://github.com/galaxyproject/ansible-galaxy-extras/blob/master/defaults/main.yml#L25

Sure, I hadn't caught it !

bgruening:
@drosofff upps sorry for that. Is this issue still relevant with @afgane suggestion? Please let me know if you have idea how to improve this in the future. In Galaxy Docker land we are always sticking to a specific commit to make sure this does not break things.

afgane:
@drosofff Let's see if that PR fixes this. 

BTW, which error in Supervisor were you seeing:
- `can't find command '/usr/sbin/automount'` → the PR should address this one
- `/usr/sbin/automount: test mount forbidden or incorrect kernel protocol version, kernel protocol version 5.00 or above required` → this on is fixed by running the container with `--privileged` flag)
[...]

PR #165, opened 2017-08-09
Move CVMFS setup before supervisor

afgane:
Otherwise, Supervisor reports `can't find command '/usr/sbin/automount'`

Should address https://github.com/galaxyproject/ansible-galaxy-extras/issues/164
If the supervisor process status is still `FATAL` at the end of a build, it always clears for me after a reboot.

### DIFF
--- a/tasks/main.yml
+++ b/tasks/main.yml
@@ -48,6 +48,11 @@
   tags:
    - condor
 
+- include: cvmfs_client.yml
+  when: galaxy_extras_config_cvmfs
+  tags:
+   - cvmfs
+   
 - include: supervisor.yml
   when: galaxy_extras_config_supervisor
   tags:
@@ -72,8 +77,3 @@
   when: galaxy_extras_config_startup
   tags:
    - startup
-
-- include: cvmfs_client.yml
-  when: galaxy_extras_config_cvmfs
-  tags:
-   - cvmfs


drosofff:
still
```
root@vm0010:~# supervisorctl status
autofs                           FATAL      Exited too quickly (process log may have details)
cron                             STOPPED    Not started
galaxy:galaxy_web                RUNNING    pid 22197, uptime 0:08:30
```
with
```
root@vm0010:~# cat /var/log/supervisor/autofs-std*
/usr/sbin/automount: program is already running.
/usr/sbin/automount: program is already running.
/usr/sbin/automount: program is already running.
/usr/sbin/automount: program is already running.
```
It's true that a reboot fixes the FATAL states.

Yet, the GalaxyKickStart test still fails. But I think it's rather a test issue with container manipulation problems. https://travis-ci.org/ARTbio/GalaxyKickStart/builds/263008332#L1362

Anyway, `galaxy_extras_config_cvmfs: false` in `group_vars/all` is a satisfactory turn around for the moment.

drosofff:
> Yet, the GalaxyKickStart test still fails. But I think it's rather a test issue with container manipulation problems. https://travis-ci.org/ARTbio/GalaxyKickStart/builds/263008332#L1362

OK, I think I figure out the reason: mount is forbidden inside containers. It's a test problem, not a code issue

bgruening:
@drosofff are you running the container with --privileged?

drosofff:
@bgruening no were are not

PR #166, opened 2017-08-15
added installation of conditional dependencies

rhpvorderman:
This is mainly important for the docker-galaxy-stable image. It will have ldap and other python dependencies installed by default if this request is pulled. We could drop some of the dependencies to save space and let the startup script of #161 handle that in case someone needs an exotic dependency.

### DIFF
--- a/defaults/main.yml
+++ b/defaults/main.yml
@@ -12,6 +12,7 @@ galaxy_extras_config_condor_docker: false
 galaxy_extras_config_k8_jobs: false
 galaxy_extras_config_supervisor: true
 galaxy_extras_config_galaxy_root: true
+galaxy_extras_config_galaxy_extra_dependencies: true
 galaxy_extras_config_galaxy_job_metrics: true
 galaxy_extras_config_uwsgi: true
 galaxy_extras_config_ie_proxy: true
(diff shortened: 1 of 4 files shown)


bgruening:
Can we add a complete new task for this and `tag` it so that we can enable, disable it conveniently?

bgruening:
ping @rhpvorderman 
Also watchdog is now part of the conditional dependencies: https://github.com/galaxyproject/galaxy/pull/4617

rhpvorderman:
The installation is now enabled or disabled by a variable and a tag.

defaults/main.yml
```diff
@@ -12,6 +12,7 @@ galaxy_extras_config_condor_docker: false
 galaxy_extras_config_k8_jobs: false
 galaxy_extras_config_supervisor: true
 galaxy_extras_config_galaxy_root: true
+galaxy_extras_config_galaxy_extra_dependencies: true
```
bgruening: Should this be false by default?

bgruening:
ping @jmchilton

defaults/main.yml
rhpvorderman: Best practice would indeed be to set it to false as it is not required for operation.

jmchilton:
Looks awesome - sorry for the delay. Thanks so much.

PR #167, opened 2017-08-24
Add IU and TACC servers to cvmfs_server_urls

chambm:
PSU server is down at the moment, adding all servers should improve robustness (when I developed this I mistakenly thought that each server was responsible for one repository (test, main, data) of the galaxyproject.org set).

### DIFF
--- a/defaults/main.yml
+++ b/defaults/main.yml
@@ -281,7 +281,7 @@ cvmfs_keys:
 cvmfs_server_urls:
   - domain: galaxyproject.org
     urls:
-      - "http://cvmfs1-psu0.galaxyproject.org/cvmfs/@fqrn@"
+      - "http://cvmfs1-iu0.galaxyproject.org/cvmfs/@fqrn@;http://cvmfs1-tacc0.galaxyproject.org/cvmfs/@fqrn@;http://cvmfs1-psu0.galaxyproject.org/cvmfs/@fqrn@"
 
 cvmfs_repositories:
   - repository: data.galaxyproject.org


chambm:
First someone needs to put up some European mirrors. ;) But once there are regional mirrors, then we can add them to the list and let CVMFS handle the logic to dynamically select the best ones based on either round trip time or geographical proximity:

> It is recommended to adjust the order of Stratum 1 servers so that the closest servers are used with priority. This can be done automatically by using geographic ordering. Alternatively, for roaming clients (clients not using a proxy server), the Stratum 1 servers can be automatically sorted according to round trip time by cvmfs_talk host probe (see Auxiliary Tools). Otherwise, the proxy server would invalidate round trip time measurement.
> 
> Ordering of Servers according to Geographic Proximity
> CernVM-FS Stratum 1 servers provide a RESTful service for geographic ordering. Clients can request http://<HOST>/cvmfs/<FQRN>/api/v1.0/geo/<proxy_address>/<server_list> The proxy address can be replaced by a UUID if no proxies are used, and the CernVM-FS client does that if there are no regular proxies. The server list is comma-separated. The result is an ordered list of indexes of the input host names. Use of this API can be enabled in a CernVM-FS client with CVMFS_USE_GEOAPI=yes. That will geographically sort both the servers set by CVMFS_SERVER_URL and the fallback proxies set by CVMFS_FALLBACK_PROXY.

bgruening:
> First someone needs to put up some European mirrors.

We are closer then you might think :)

PR #169, opened 2017-10-18
fix for hard-coded uwsgi ansible variable uwsgi_port

luke-c-sargent:
Hello; this is a fix for a situation where a user [specifies their own uwsgi_port in this role's default variables](https://github.com/galaxyproject/ansible-galaxy-extras/blob/7ab0cad82e71ed71678aa0443b5c7d9808439ed4/defaults/main.yml#L120). 

Port 4001 was hardcoded into the `supervisor.conf` template, so galaxy_web uwsgi instances were listening on that port, however `nginx.conf` [correctly picked up the new port](https://github.com/galaxyproject/ansible-galaxy-extras/blob/4b6344450ccd50d0b26806f3e0c1ed39b45aaeee/templates/nginx_uwsgi.conf.j2#L3) and was `proxy_pass`ing requests to the specified port.

With this minor alteration, nginx and uwsgi once again have a harmonious relationship.

*note: i wasn't able to find any other places where the port was hard-coded in, or any other hard-coded ports that could benefit from being turned in to variables, but am happy to make those alterations if there are some i've missed*

### DIFF
--- a/templates/supervisor.conf.j2
+++ b/templates/supervisor.conf.j2
@@ -102,9 +102,9 @@ priority        = 200
 [program:galaxy_web]
 {% if galaxy_uwsgi|bool %}
 {% if galaxy_uwsgi_static_conf|bool %}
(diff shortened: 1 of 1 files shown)


afgane:
Good addition.

PR #172, opened 2017-10-19
Initial defaults for missing k8s plugin vars

pcm32:
This PR adds some missing k8s plugin vars. As discussed in https://github.com/galaxyproject/galaxy-kubernetes/issues/2#issuecomment-336390996 .

Not sure if this might break something on older galaxy versions were the k8s runner wasn't aware of these additional vars. Please advise.

### DIFF
--- a/templates/job_conf.xml.j2
+++ b/templates/job_conf.xml.j2
@@ -31,6 +31,11 @@
             <param id="k8s_use_service_account">true</param>
             <param id="k8s_persistent_volume_claim_name">galaxy-web-claim0</param>
             <param id="k8s_persistent_volume_claim_mount_path">/export</param>
+            <param id="k8s_namespace">default</param>
+            <param id="k8s_supplemental_group_id">0</param>
+            <param id="k8s_fs_group_id">0</param>
+            <param id="k8s_pull_policy">IfNotPresent</param>
+            <param id="k8s_pod_retrials">1</param
 {% if galaxy_minimum_version >= "17.09" %}
             <param id="enabled" from_environ="GALAXY_RUNNERS_ENABLE_K8">true</param>
 {% endif %}
--- a/templates/job_conf.xml.j2
+++ b/templates/job_conf.xml.j2
@@ -28,14 +28,14 @@
 {% if galaxy_extras_config_k8_jobs %}
         <plugin id="k8" type="runner" load="galaxy.jobs.runners.kubernetes:KubernetesJobRunner">
             <!-- We are inside of Kubernetes so use this. -->
-            <param id="k8s_use_service_account">{{ use_service_account }}</param>
-            <param id="k8s_persistent_volume_claim_name">{{ galaxy_k8s_jobs_persistent_volume_claim_name }}</param>
-            <param id="k8s_persistent_volume_claim_mount_path">{{ galaxy_k8s_jobs_persistent_volume_claim_mount_path }}</param>
-            <param id="k8s_namespace">{{ galaxy_k8s_jobs_namespace }}</param>
-            <param id="k8s_supplemental_group_id">{{ galaxy_k8s_jobs_supplemental_group_id }}</param>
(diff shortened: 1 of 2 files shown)


pcm32:
Will add ansible placeholders soon for values.

pcm32:
@bgruening where is the logic that translates ENV vars into ansible variables? Or do I need to add it to the [k8s tasks](https://github.com/galaxyproject/ansible-galaxy-extras/blob/master/tasks/k8.yml)? Thanks!

bgruening:
Try to add `from_environ="GALAXY_RUNNERS_ENABLE_PBS"` or similar to your params in your job_conf. There are a few examples in the file already.

pcm32:
Ahh, I see, so we skip ansible variables entirely?

bgruening:
We could. But if you want to set hardcoded default variables during build-time, they should stay.

bgruening:
I would leave them. This PR looks good.

bgruening:
You removed the WIP, but did not include the from_environ ...

pcm32:
Ahhh, sorry, I thought you had merged... will do. Give me a sec.

pcm32:
I was tempted to change 
`GALAXY_RUNNERS_ENABLE_K8` for 
`GALAXY_RUNNERS_ENABLE_K8S` 
but that probably requires changes elsewhere and in the container, can be part later of another PR.

Will remove [WIP], let me know if we need to change anything.

bgruening:
Need to jump in a train. But will get to this today. Thanks Pablo!

templates/job_conf.xml.j2
```diff
@@ -28,9 +28,14 @@
 {% if galaxy_extras_config_k8_jobs %}
         <plugin id="k8" type="runner" load="galaxy.jobs.runners.kubernetes:KubernetesJobRunner">
             <!-- We are inside of Kubernetes so use this. -->
-            <param id="k8s_use_service_account">true</param>
-            <param id="k8s_persistent_volume_claim_name">galaxy-web-claim0</param>
-            <param id="k8s_persistent_volume_claim_mount_path">/export</param>
+            <param id="k8s_use_service_account" from_environ="GALAXY_RUNNERS_K8S_USE_SERVICE_ACCOUNT">{{ use_service_account }}</param>
+            <param id="k8s_persistent_volume_claim_name" from_environ="GALAXY_RUNNERS_K8S_PVC_NAME">{{ galaxy_k8s_jobs_persistent_volume_claim_name }}</param>
+            <param id="k8s_persistent_volume_claim_mount_path" from_environ="GALAXY_RUNNERS_K8S_PVC_MOUNT_PATH">{{ galaxy_k8s_jobs_persistent_volume_claim_mount_path }}</param>
```
bgruening: Wouldn't it be better to have the same ENV var name as the ID, but prefixed with `GALAXY_RUNNERS_K8S_`

templates/job_conf.xml.j2
```diff
@@ -28,9 +28,14 @@
...
+            <param id="k8s_namespace" from_environ="GALAXY_RUNNERS_K8S_NAMESPACE">{{ galaxy_k8s_jobs_namespace }}</param>
+            <param id="k8s_supplemental_group_id" from_environ="GALAXY_RUNNERS_K8S_SUPP_GROUP_ID">{{ galaxy_k8s_jobs_supplemental_group_id }}</param>
```
bgruening: here as well?

templates/job_conf.xml.j2
pcm32: yes, I thought the same, but they were too long for the PVCs. Do you prefer these changed?

templates/job_conf.xml.j2
pcm32: ^^^ idem

templates/job_conf.xml.j2
afgane: I think keeping them consistent is better, despite the length. That way it's one rule vs opening door for other deviations.

bgruening:
Thanks @pcm32!

PR #175, opened 2017-11-05
Adjust slurm.conf for Ubuntu 16.04 Xenial

drosofff:
*Issue:*

Whereas the role works fine with Ubuntu 14.04, queuing of jobs in slurm cluster is suboptimal in Ubuntu 16.04 Xenial.

Namely, jobs are put in slurm wait queue with `Resources` in `NODELIST(REASON)` and are only launched one at a time, even if CPUs are available.

Thus, with two fastqc jobs launched under Galaxy + Ubuntu 14.04:

```
             JOBID PARTITION     NAME     USER ST       TIME  NODES NODELIST(REASON)
                 9     debug g12_fast   galaxy  R       0:15      1 ubuntu-14-04
                10     debug g13_fast   galaxy  R       0:10      1 ubuntu-14-04
```
(OK)

with two fastqc jobs launched under Galaxy + Ubuntu 16.04:

```
             JOBID PARTITION     NAME     USER ST       TIME  NODES NODELIST(REASON)
                 3     debug g31_fast   galaxy PD       0:00      1 (Resources)
                 2     debug g30_fast   galaxy  R       0:40      1 ubuntu-16-04
```
(Not OK, job 3 is unnecessarily stalled)

*Fix:*
Replace `SelectTypeParameters=CR_Core_Memory` with `SelectTypeParameters=CR_CPU`in `configure_slurm.py.j2`

Now, with two fastqc jobs launched under Galaxy + Ubuntu 14.04:

```
             JOBID PARTITION     NAME     USER ST       TIME  NODES NODELIST(REASON)
                 9     debug g12_fast   galaxy  R       2:56      1 ubuntu-14-04
                10     debug g13_fast   galaxy  R       2:51      1 ubuntu-14-04
```
(Still OK, parameter change does not impair the role)

with two fastqc jobs launched under Galaxy + Ubuntu 16.04:

```
             JOBID PARTITION     NAME     USER ST       TIME  NODES NODELIST(REASON)
                 2     debug g30_fast   galaxy  R       3:42      1 ubuntu-16-04
                 3     debug g31_fast   galaxy  R       0:48      1 ubuntu-16-04
```
(Fixed, the job 3 is launched in parallel with job 2, using all CPUs available)


bgruening:
@drosofff but this change will not track the memory usage right? How much memory is available? Maybe the memory consumption is calculated differently with a new SLURM version?

drosofff:
@bgruening that is right, memory usage is not expected to by tracked with SelectTypeParameters=CR_Core, from what I can read in https://slurm.schedmd.com/cons_res_share.html, `Memory management` section.

Yet, if we speak of a single node cluster (standard case with these ansible roles), I don't really catch the purpose of CR_Core_Memory or CR_CPU_Memory. But I am reaching my limits in cluster management...

It is likely that something has changed in slurm in 16.04 Xenial and I really don't understand why the original setting CR_Core_Memory has the afore described behavior in this OS environment

In any case, I have tested the 4 possibilities under 16.04 Xenial and from CR_Core_Memory, CR_CPU_Memory, CR_core and CR_CPU, only the CR_CPU option allows consuming the submitted jobs as expected.

Maybe @natefoo can give some input on this ? I have been looking for hints on the issue for a while on the net, but could not find anything really inspiring...

templates/configure_slurm.py.j2
```diff
@@ -58,7 +58,7 @@ FastSchedule=1
 SchedulerType=sched/backfill
 SchedulerPort=7321
 SelectType=select/cons_res
-SelectTypeParameters=CR_Core_Memory
+SelectTypeParameters=CR_CPU
```
mvdbeek: We could make this an ansible variable for the time being, right ?

natefoo:
Any 4 of CR_CPU, CR_CPU_Memory, CR_Core, or CR_Core_Memory should allow multiple jobs to run. Can you check the output of `sinfo -Nel` with one of the _Memory select types enabled?

templates/configure_slurm.py.j2
drosofff: Sure.

I'll check as suggested by @natefoo

drosofff:
@natefoo 

with SelectTypeParameters=CR_Core_Memory
```
root@slurm-test-1604:~# sinfo -Nel
Wed Nov 15 22:03:12 2017
NODELIST         NODES PARTITION       STATE CPUS    S:C:T MEMORY TMP_DISK WEIGHT FEATURES REASON
slurm-test-1604      1    debug*       mixed    2    2:1:1   7477        0      1   (null) none
```
```
root@slurm-test-1604:~# squeue
             JOBID PARTITION     NAME     USER ST       TIME  NODES NODELIST(REASON)
                 3     debug g2_uploa   galaxy PD       0:00      1 (Resources)
                 2     debug g1_uploa   galaxy  R       2:58      1 slurm-test-1604
```
Whereas right after SelectTypeParameters=CR_Core and restarting slurm:
```
root@slurm-test-1604:~# sinfo -Nel
Wed Nov 15 22:07:11 2017
NODELIST         NODES PARTITION       STATE CPUS    S:C:T MEMORY TMP_DISK WEIGHT FEATURES REASON
slurm-test-1604      1    debug*   allocated    2    2:1:1   7477        0      1   (null) none
root@slurm-test-1604:~# squeue
             JOBID PARTITION     NAME     USER ST       TIME  NODES NODELIST(REASON)
                 2     debug g1_uploa   galaxy  R       8:21      1 slurm-test-1604
                 3     debug g2_uploa   galaxy  R       0:57      1 slurm-test-1604
```

natefoo:
Interesting, Slurm sees 2 CPUs and sees that one is free, hence the `mixed` node state. Is it possible that the job's memory requirement is preventing it from running?

drosofff:
> Is it possible that the job's memory requirement is preventing it from running?

I find it unlikely: This bug is not new to me and I have been working on it from time to time using a new server we have under ubuntu 16.04 and with 384 Gb RAM. Same problem. Moreover, among the numerous tests I made with VM I have also cases where the tests jobs was on purpose requiring more memory than can be supplied by the system: in this case the (Reason) turns to (Memory), as expected. Finally, I remind you that it is completely release specific: exactly the same slurm.conf under 14.04 works fine with the playbook default settings, and the same VM specification (GCE 2 core, 7.5 Gb RAM).
I realize that there is something that I did not tested: a VM in amazon. But as I reproduce the same bug with a DELL Bare metal server, I am expecting the same bug. I'll try it through.
Thanks for the help anyway.

natefoo:
Ok. We ran Slurm 15.08 for a while before upgrading to a newer version and used `CR_CPU_Memory`, so it should work. I can keep digging if you'd like: if yes, then please post your `slurm.conf` as well as the outputs of `scontrol show node slurm-test-1604`, `scontrol show partition debug` and `scontrol show job N` for each N = job id of running/pending job.

drosofff:
OK. You'll find the posts in gist https://gist.github.com/drosofff/248377376b3829c25eef17a5069f3c75
Thanks !

natefoo:
Okay, so your node has 7477 MB of memory. When the first job runs, it does so without `--mem` or `--mem-per-cpu`, so by default, it is allocated all of the memory on the node. This prevents the second job from running. There are multiple possible solutions:

1. Set `DefMemPerNode` or `DefMemPerCPU` on the partition:

```
PartitionName=debug Nodes=slurm-test-1604 Default=YES MaxTime=INFINITE State=UP Shared=YES DefMemPerCPU=3737
```

2. Use `--mem` or `--mem-per-cpu` when submitting.

Ideally there would be a way to specify `DefMemPerCPU=<RealMemory/CPUs>` but no such option exists.

drosofff:
Thank @natefoo . For complete clarification: do you understand by it behaves differently in U14.04 and U16.04 ? Is it an issue or a new feature/evolution of the software ?

Do you agree that in its current state, galaxy-extra is suboptimal (requires advanced tuning) for installation in Ubuntu 16.04 ?

natefoo:
I'm not sure why it worked in 14.04 - as far as I can tell, for memory settings to work, you always had to specify either a default memory allocation on the partition or use `--mem`. It's possible that in 14.04's version (2.6.5), the default action if neither was specified was for the memory resource to be ignored? Without testing, I don't know why it works differently.

I agree it's suboptimal. My recommendation would be to set `DefMemPerCPU=<RealMemory/CPUs>` when creating `slurm.conf`. I see the memory value comes from `psutil` now but we should be able to use `slurmd`'s value to ensure that the value of `DefMemPerCPU * CPUs` does not exceed `RealMemory`, e.g.:

```sh-session
nate@weyerbacher% slurmd -C | sed 's/.*\( \+\|^\)RealMemory=\([0-9][0-9]*\)\( \|$\).*/\2/p;d'
32084
% slurmd -C | sed 's/.*\( \+\|^\)CPUs=\([0-9][0-9]*\)\( \|$\).*/\2/p;d'
8
% echo $(($(slurmd -C | sed 's/.*\( \+\|^\)RealMemory=\([0-9][0-9]*\)\( \|$\).*/\2/p;d') / $(slurmd -C | sed 's/.*\( \+\|^\)CPUs=\([0-9][0-9]*\)\( \|$\).*/\2/p;d')))
4010
```

The `slurmd -C` output can be parsed in python instead of course.

drosofff:
OK, thanks a lot for the explanations @natefoo . I'll try to test again and incorporate your suggestions in the playbook

drosofff:
What do you think of that @natefoo ? Note that DefMemPerCPU does not behave with slurm 2.x as it does with slurm 15.x.

tasks/slurm.yml
```diff
@@ -47,6 +47,14 @@
 - name: Setup tmp area for slurm.
   file: path=/tmp/slurm state=directory owner={{ galaxy_user_name }} group={{ galaxy_user_name }}
 
+- name: Collect Memory/cpu metric for slurm.conf
+  shell: echo $(($(slurmd -C | sed 's/.*\( \+\|^\)RealMemory=\([0-9][0-9]*\)\( \|$\).*/\2/p;d') / $(slurmd -C | sed 's/.*\( \+\|^\)CPUs=\([0-9][0-9]*\)\( \|$\).*/\2/p;d')))
```
natefoo: The memory amounts as reported by `slurmd -C` and by `psutil` might differ. If the `psutil` amount is smaller than `slurmd -C` that is not a problem, but if it's larger, that would prevent 1 slot on the node from being usable. I would suggest collecting both RealMemory and CPUs from `slurmd -C`, e.g.:

```yaml
- name: Collect slurmd -C output
  command: slurmd -C
  register: slurmd_c_out

- name: Make slurmd -C output a single line
  set_fact:
    slurmd_c: "{{ slurmd_c_out.stdout | regex_replace('\\n', ' ') }}"

- name: Register Slurm-reported facts for configure_slurm.py.j2
  set_fact:
        slurm_real_memory: "{{ slurmd_c | regex_replace('.*(?:^|\\s)RealMemory=(\\d+)(?:$|\\s).*', '\\1') }}"
        slurm_cpus: "{{ slurmd_c | regex_replace('.*(?:^|\\s)CPUs=(\\d+)(?:$|\\s).*', '\\1') }}"

- name: Calculate memory per CPU for configure_slurm.py.j2
    - set_fact:
        slurm_memory_per_cpu: "{{ slurm_real_memory | int // slurm_cpus | int }}"
```

And then use `{{ slurm_real_memory }}` and `{{ slurm_cpus }}` in place of the psutil-collected variables in configure_slurm.py.j2

drosofff: OK, I see and agree, thanks !

drosofff:
@natefoo More importantly, the config accommodates much better slurm v15 found in Ubuntu Xenial. Thanks a lot for the help !

templates/configure_slurm.py.j2
```diff
@@ -98,7 +99,9 @@ def main():
         "user": environ.get('SLURM_USER_NAME', '{{ galaxy_user_name }}'),
         "cpus": environ.get("SLURM_CPUS", cpus),
         "partition_name": environ.get('SLURM_PARTITION_NAME', 'debug'),
-        "memory": environ.get("SLURM_MEMORY", int(mem / (1024 * 1024)))
+        "memory": environ.get("SLURM_MEMORY", int(mem / (1024 * 1024))),
+        "memory_per_cpu": environ.get("SLURM_MEMPERCPU",
+                                      int(mem / (1024 * 1024)) / cpus)
```
natefoo: These 3 lines can be removed now, as well as the `cpus` key and the entire `try` block above, I believe.

drosofff: sure !

natefoo:
👍 thanks @drosofff!

templates/configure_slurm.py.j2
```diff
@@ -58,7 +58,7 @@ FastSchedule=1
 SchedulerType=sched/backfill
 SchedulerPort=7321
 SelectType=select/cons_res
-SelectTypeParameters=CR_Core_Memory
+SelectTypeParameters={{ slurm_SelectTypeParameters }}
```
mvdbeek: I know I proposed adding ansible vars in here (it's a .j2 file after all), but I think this will set variables at build time for @bgruening's docker image which may or may not be good defaults. Should we move the calculation into the `main` function of this script ? That way this gets adjusted when you start up the docker image without needing to let ansible touch this file.

mvdbeek: That is if @bgruening is still using this, I didn't follow the development closely.

drosofff: Well, the change is not really needed anymore with the @natefoo explanations. We may go back harmlessly to hard-coded CR_Core_Memory.

bgruening: Yeah, we are still using this. Thanks @mvdbeek for pointing this out.
But I guess we really can redo these changes now.

drosofff: No problem, I'll reverse to `SelectTypeParameters=CR_Core_Memory`
Thank you all for your help

templates/configure_slurm.py.j2
```diff
@@ -76,29 +76,19 @@ SlurmctldDebug=3
 #SlurmctldLogFile=
 SlurmdDebug=3
 #SlurmdLogFile=
-NodeName=$hostname CPUs=$cpus RealMemory=$memory State=UNKNOWN
-PartitionName=$partition_name Nodes=$hostname Default=YES MaxTime=INFINITE State=UP Shared=YES
+NodeName=$hostname CPUs={{ slurm_cpus }} RealMemory={{ slurm_real_memory }} State=UNKNOWN
```
bgruening: @drosofff this is still set during build time with no way to set it during runtime, right?
Wouldn't it be enough to replace the psutil call with the `slurmd -C` magic and configure everything during runtime as before?

drosofff: Sure, the more satisfied people are, the happier I am ;-)

drosofff:
@bgruening how about that ?

drosofff:
Ping @bgruening

bgruening:
Seems to work. Thanks @drosofff!

PR #177, opened 2017-11-14
fix ie_proxy.yml

drosofff:
- As mentioned in [RTD](https://docs.galaxyproject.org/en/master/admin/interactive_environments.html), : 
```
if you have NodeJS installed under Ubuntu, it often installs to /usr/bin/nodejs, whereas npm expects it to be /usr/bin/node. You will need to create that symlink yourself.
```

- I find that Install proxy dependencies task as `galaxy` user fails. Is there any reason to do so ? My proposed cut allows ansible to install the nodejs proxy in all cases

### DIFF
--- a/tasks/ie_proxy.yml
+++ b/tasks/ie_proxy.yml
@@ -1,7 +1,7 @@
 - name: Install web related system packages
   apt: pkg={{ item }} state={{ galaxy_extras_apt_package_state }}
   with_items:
-    - nodejs
+    - nodejs-legacy
     - npm
   when: galaxy_extras_install_packages
 
--- a/tasks/ie_proxy.yml
+++ b/tasks/ie_proxy.yml
@@ -12,6 +12,8 @@
 
 - name: "Install proxy dependencies."
   shell: "cd {{ galaxy_server_dir }}/lib/galaxy/web/proxy/js && npm install"
+  become: True
+  become_user: "{{ galaxy_user_name }}"
 
 - name: "Install Juypter container."
   shell: "docker pull {{ galaxy_extras_ie_jupyter_image }}"
--- a/tasks/ie_proxy.yml
+++ b/tasks/ie_proxy.yml
@@ -9,6 +9,7 @@
   file:
     src: /usr/bin/nodejs
     dest: /usr/bin/node
+    state: link
 
 - name: "Install proxy dependencies."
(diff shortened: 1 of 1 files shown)


tasks/ie_proxy.yml
```diff
@@ -5,10 +5,13 @@
...
   when: galaxy_extras_install_packages
 
+- name: Symlinks node to nodejs
+  file:
+    src: /usr/bin/nodejs
+    dest: /usr/bin/node
+
 - name: "Install proxy dependencies."
   shell: "cd {{ galaxy_server_dir }}/lib/galaxy/web/proxy/js && npm install"
-  become: True
```
bgruening: Do you know why this fails on a new Ubunut?

drosofff: I need to dig again in that to answer...
In the mean time, do you know why this task has to be performed with become_user: galaxy ?

bgruening: Its installing libraries into the Galaxy source tree, so I figured it would be correct to have it own as galaxy user. I'm not sure if it really needs to be owed by Galaxy. But probably better then root.

drosofff: @bgruening My bad Bjorn, I retested with become / become_user (galaxy) and it works too. My dev instance was probably deriving with non-ansible settings... I'll roll back this tomorrow

drosofff: > Do you know why this fails on a new Ubunut?

By the way, `npm install` by galaxy raises the same problem in 14.04 and 16.04 (fatal error at first `supervisorctl start galaxy:proxy_node_js`), but as mentioned elsewhere, I found a turn around in the GKS playbook

tasks/ie_proxy.yml
```diff
@@ -5,6 +5,11 @@
     - npm
   when: galaxy_extras_install_packages
 
+- name: Symlinks node to nodejs
+  file:
+    src: /usr/bin/nodejs
+    dest: /usr/bin/node
```
bgruening: Should we add `state: link` here?

drosofff: sure, it saves space !

drosofff: ... and you're right it does what it says it does !

bgruening:
Cool, thanks!

PR #178, opened 2017-11-20
Fixes abbreviation used for k8s, so that it is coherent everywhere.

pcm32:
This PR fixes the abbreviation use of k8s, which being incoherently used as k8 and k8s at different parts was inducing me to all sort of silly mistakes when deploying. I'm aware that this requires changes at the compose side, but I thought it would be better to do this now at this point given that less people are using it.

I will submit another PR at compose side.


jmchilton:
Sorry about that - definitely worth breaking backward compatibility to make this consistent - I really messed up here.

pcm32:
Thanks for the support John, no worries! Given that, I will proceed with the PR at the other side.

pcm32:
Here is the [other PR](https://github.com/bgruening/docker-galaxy-stable/pull/406), in the waiting for a commit sha at this side.

bgruening:
Thanks @pcm32!

PR #179, opened 2017-11-28
Add new CVMFS mirror, the first European one!

bgruening:


afgane:
Did you happen to write any docs on how to setup a mirror?

bgruening:
@afgane kind of ;) We have a huge email thread ;) 
We will setup one mirror in Freiburg as well and will document it in a better way.

afgane:
Validated working, even from the US ;)

PR #180, opened 2017-12-14
Change signal for stopping galaxy:galaxy_web to stopsignal=QUIT

drosofff:
In response to issue discussed in https://github.com/galaxyproject/ansible-galaxy-tools/issues/48,
here we propose to increase the strength of the signal termination sent by supervisor to uwsgi when stopping or restarting the galaxy:web_galaxy service.
This makes more robust the management of this service by a handler discussed in the issue

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,0 +1,6 @@
+.idea/
+.vagrant/
+*.retry
+.DS_Store
+*.pyc
+docs/html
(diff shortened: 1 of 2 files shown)


drosofff:
ping @everyone

bgruening:
Thanks @drosofff!

PR #181, opened 2018-01-01
checking for a running pg instance

bgruening:


mvdbeek:
It's not the exact same thing, but would [pg_isready](https://www.postgresql.org/docs/9.3/static/app-pg-isready.html) work instead of the custom script ?

bgruening:
Yes we could, but I was not sure if we can assume >=9.3, can we?

bgruening:
@mvdbeek have you seen my last comment? Are you ok with this PR?

mvdbeek:
Sure, no problem.

Issue #183, opened 2018-01-26
nginx delegated downloads


drosofff:
Before I forget it, I raise the issue:

In
https://github.com/galaxyproject/ansible-galaxy-extras/blob/master/templates/nginx.conf.j2#L121

I don't think
```
        # delegated downloads
        location {{ nginx_galaxy_location }}/_x_accel_redirect {
```
is correct.

In agreement with https://docs.galaxyproject.org/en/latest/admin/special_topics/nginx.html#sending-files-using-nginx, it should be
```
        # delegated downloads
        location /_x_accel_redirect {
```
with or without nginx prefixing

drosofff:
Fixed in PR #185

PR #185, opened 2018-02-02
Change location {{ nginx_galaxy_location }}/_x_accel_redirect {

drosofff:
to     location /_x_accel_redirect/ {

as explained in #183, nginx delegated downloads should not be prefixed even with prefixed galaxy frontend. In its current form, downloads fail for prefixed server instances.

### DIFF
--- a/templates/nginx.conf.j2
+++ b/templates/nginx.conf.j2
@@ -119,7 +119,7 @@ http {
         }
 
         # delegated downloads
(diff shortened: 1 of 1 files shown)


drosofff:
And for more extensive testing, see https://travis-ci.org/ARTbio/GalaxyKickStart/jobs/336520437

bgruening:
@drosofff thanks!

PR #187, opened 2018-02-07
post-start-actions script placeholder

### LINKED ISSUE #184: Post-startup script
Hi there! I would like to have an option to provide an optional post-flight (or post startup script), which admins could use for all different things related to the configuration of the live instance (or other things that require a living galaxy instance, API accessible).

I would suggest adding some logic like:

```
if [[ -x $POST_START_SCRIPT ]]
   then
          # ephemeris installed already, maybe check for this
          galaxy-wait -g http://127.0.0.1 -v --timeout 120 > {{ galaxy_log_dir }}/post_start_actions.log && \
          $POST_START_SCRIPT >> {{ galaxy_log_dir }}/post_start_actions.log &
          # So this will wait for galaxy to start, without delaying the execution of the last tails commands that go after the user setup section.
fi
```

after or within https://github.com/galaxyproject/ansible-galaxy-extras/blob/24ad229a64dab6f1bf15bc825ee3cbfe6b018fbf/templates/startup.sh.j2#L426

as you might need user authentication and API Key.

Then admins would have the liberty to set a $POST_START_SCRIPT and do things like adding users (for a course for instance), adding workflows, setting quotas, etc.

I realise that this opens the door for trouble, but then is a price that comes with flexibility. For added security, we could have a standard location for the post start script, and refer to it directly (and then admins have to create the file with proper content).

Would this be well received as a contribution? Alternatively I can just PR the exact thing I need to have optionally done (which is to add workflows to a user and make them publishable).

pcm32:
This PR adds the ability for admins to add an script with logic that should be executed once the galaxy instance is running. `galaxy-wait` is part of ephemeris which is installed in the docker-stable compose images that are relevant to the `startup` script.

I modified slightly the way that the create-user is invoked, to better comply with its usage of admin email and admin user. Previously the email wasn't being used, which risks the admin not being blessed as such, but we can discuss this if needed. 

I have tested the functionality to work with my real use case.

This resolves #184

### DIFF
--- a/templates/startup.sh.j2
+++ b/templates/startup.sh.j2
@@ -428,6 +428,19 @@ if [ "x$GALAXY_DEFAULT_ADMIN_USER" != "x" ]
     then
         echo "Creating admin user $GALAXY_DEFAULT_ADMIN_USER with key $GALAXY_DEFAULT_ADMIN_KEY and password $GALAXY_DEFAULT_ADMIN_PASSWORD if not existing"
         python /usr/local/bin/create_galaxy_user.py --user "$GALAXY_DEFAULT_ADMIN_USER" --password "$GALAXY_DEFAULT_ADMIN_PASSWORD" -c "$GALAXY_CONFIG_FILE" --key "$GALAXY_DEFAULT_ADMIN_KEY"
+
+	# If there is a need to execute actions that would require a live galaxy instance, such as adding workflows, setting quotas, adding more users, etc.
+	# then place a file with that logic named post-start-actions.sg on the Galaxy working directory, it should have access to all environment variables 
+	# visible here.
+	# The file needs to be executable (chmod a+x post-start-actions.sh)
+        if [ -x ./post-start-actions.sh ]
+            then
+	       # uses ephemeris, present in docker-galaxy-stable, to wait for the local instance
+	       galaxy-wait -g http://127.0.0.1 -v --timeout 120 > {{ galaxy_log_dir }}/post-start-actions.log &&
+	       ./post-start-actions.sh >> {{ galaxy_log_dir }}/post-start-actions.log &    
+	fi
+
+
 fi
 
 # Enable verbose output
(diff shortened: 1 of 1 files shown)


templates/startup.sh.j2
```diff
@@ -427,7 +427,20 @@ fi
 if [ "x$GALAXY_DEFAULT_ADMIN_USER" != "x" ]
     then
         echo "Creating admin user $GALAXY_DEFAULT_ADMIN_USER with key $GALAXY_DEFAULT_ADMIN_KEY and password $GALAXY_DEFAULT_ADMIN_PASSWORD if not existing"
-        python /usr/local/bin/create_galaxy_user.py --user "$GALAXY_DEFAULT_ADMIN_USER" --password "$GALAXY_DEFAULT_ADMIN_PASSWORD" -c "$GALAXY_CONFIG_FILE" --key "$GALAXY_DEFAULT_ADMIN_KEY"
+        python /usr/local/bin/create_galaxy_user.py --user "$GALAXY_DEFAULT_ADMIN_EMAIL" --password "$GALAXY_DEFAULT_ADMIN_PASSWORD" \
+		-c "$GALAXY_CONFIG_FILE" --username "$GALAXY_DEFAULT_ADMIN_USER" --key "$GALAXY_DEFAULT_ADMIN_KEY"
+	# If there is a need to execute actions that would require a live galaxy instance, such as adding workflows, setting quotas, adding more users, etc.
+	# then place a file with that logic named post-start-actions.sg on the Galaxy working directory, it should have access to all environment variables
```
bgruening: `s/sg/sh/`

bgruening: Can we call Galaxy working dir, `the /export/` directory or similar?

pcm32: Thanks for looking into it. I have added a commit with these changes.

bgruening:
Thanks :)

PR #189, opened 2018-02-14
Feature/isatools-lite installation

pcm32:
This PR adds an isatools-lite optional install managed by ansible to support the functionality of the isa-tab and isa-json composite datatypes recently contributed by PhenoMeNal. This will be executed in the compose galaxy-init container provided that a build argument `--build-arg ISATOOLS_LITE_INSTALL=True` is given on docker build.

It also pins uwsgi to version 2.0.15 as a new version 2.0.16 came for which there is no wheel, producing failures when building containers.

### DIFF
--- a/tasks/isatools.yml
+++ b/tasks/isatools.yml
@@ -1,0 +1,35 @@
+# Install the unofficial isatools lite for the Galaxy ISA -tab and -json datatypes
+# Assumes pip 9
+- name: Install packages 
+  apt: pkg={{ item }} state={{ galaxy_extras_apt_package_state }}
+  with_items:
+   - git
+  when: isatools_lite_install
+
+- name: Install networkx
+  pip:
+    name: networkx
+    virtualenv: "{{ galaxy_venv_dir }}"
+    virtualenv_command: "{{ pip_virtualenv_command | default( 'virtualenv' ) }}"
+  when: isatools_lite_install
+  become: True
+  become_user: "{{ galaxy_user_name }}"
+
+- name: Install isatools-lite unofficial
+  pip:
+    name: git+https://github.com/ilveroluca/isa-rwval@feature/fewer_deps#egg=isatools
+    virtualenv: "{{ galaxy_venv_dir }}"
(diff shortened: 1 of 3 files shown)


jmchilton:
How are you installing the datatypes in Galaxy - are you pointing at a fork of Galaxy with the datatypes? If yes, can't you just add these to that Galaxy's requirements file so these are picked automatically?

pcm32:
Hi! The main issue I have is that isatools is very heavy by default on dependencies, but these dependencies are not really needed for the datatype, so we need to install without dependencies to avoid bloating stuff. But I see your point of course that this should be installed as all other Galaxy deps in the Galaxy requirements. Can you add entries into `requirements.txt` that avoid following dependencies? We are talking about ~400 MB of added dependencies. 

I expect this to be a transition installation mode, as we are working with colleagues in generating a branch of ISA that can be installed with less trouble.

pcm32:
Saw some unexpected issues... let me set this to WIP for a little while.

pcm32:
This effort is irrelevant since the ansible call used by kn makes use of netifaces on runtime.

pcm32:
Sorry, closed wrong PR!

bgruening:
Thanks @pcm32 for working on this!

pcm32:
Hi think I should be able to avoid this whole non-dependency installation soon. Maybe I'll open another PR just for pinning uwsgi

abretaud:
> Maybe I'll open another PR just for pinning uwsgi

@pcm32 if you have time, yes, it'd be great ! I need it too to fix docker image building

pcm32:
@abretaud please see PR #190 which only includes that change, hopefully it will be merged straight away by @bgruening or other admin.

abretaud:
Coooool, thanks @pcm32!

pcm32:
I have added now a PR at starforge to get the minimal ISA-tools that we are using installed:
https://github.com/galaxyproject/starforge/pull/197

That would make this PR unnecessary :-) - or simplify it a lot.

pcm32:
I don't think that this is needed any longer....

PR #190, opened 2018-02-23
pins uwsgi version to 2.0.15

pcm32:
Sorts failing build of dockerfile on compose setup.

### DIFF
--- a/tasks/uwsgi.yml
+++ b/tasks/uwsgi.yml
@@ -1,6 +1,7 @@
 - name: "Install uwsgi for galaxy"
   pip: 
     name: uwsgi
+    version: 2.0.15
     virtualenv: "{{ galaxy_venv_dir }}"
     virtualenv_command: "{{ pip_virtualenv_command | default( 'virtualenv' ) }}"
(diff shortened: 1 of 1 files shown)


bgruening:
I'm not sure why, but it can not find the uwsgi package or tries to get it from pypi :(

pcm32:
mmm... weird, I have been building for the past days, including today, with this version.... where are you getting the issue? when triggering the build of all containers together?

bgruening:
@pcm32 I was referring to the tests. But maybe there is something else fishy?

pcm32:
ahhh, I see what you mean now... funny that the test is sort of looking for 2.0.16 (or installing it) when we are actually pinning 2.0.15. There are some weird permission denied as well that I wouldn't expect (but then I don't do a lot of travis, so maybe is typical on that setup).

abretaud:
The 2.0.16 install comes from the galaxy stable image I think, I made https://github.com/bgruening/docker-galaxy-stable/pull/421

bgruening:
And green. Thanks you both. Will follow up on dev.

PR #191, opened 2018-02-25
install nodejs 9.x for latest JS packages

bgruening:


bgruening:
The old proxy JS packages disappeared to we needed to pin newer once, these newer packages now complain during installation about our old nodejs-legacy, so I ended up updating our node installation. Which is not that bad as we need it for your new client as well.

ping @jmchilton @natefoo

PR #192, opened 2018-02-25
[18.01] tracking branch for the Galaxy Docker Project v18.01

bgruening:
I will maintain this branch for some time I guess to get the Galaxy Docker Project into shape for 18.01.

Due to lack of time my primarily goal is to get the image to compile and travis to pass. A lot of cleanups can and should be happen to cut ties and remove paster everywhere.

A few changes included:

* a few old proxy JS packages (sqlite...) disappeared and we needed to pin newer once, these newer packages now complain during installation about our old nodejs-legacy, so I ended up updating our node installation. Which is not that bad as we need it for your new client as well.

* Initial changes for `galaxy.ini` to `galaxy.yml` migration. I had hoped to skip this in 18.01 and do it later this year but due to the missing `galaxy.ini` I started the migration

* skip a bioblend test for samtools conda installation (https://github.com/galaxyproject/galaxy-flavor-testing/tree/18.01) which will not be needed as soon as we get a new bioblend release

@jmchilton @natefoo @drosofff this might be interesting for you.
@drosofff I do not intend to merge this branch soon to not break you setup, but I will use this branch in the Docker project.


tasks/ie_proxy.yml
```diff
@@ -1,15 +1,28 @@
...
+    - "deb-src https://deb.nodesource.com/node_{{ nodejs_version }} {{ ansible_distribution_release }} main"
 
-- name: Symlinks node to nodejs
-  file:
-    src: /usr/bin/nodejs
-    dest: /usr/bin/node
-    state: link
+- name: Install nodejs (npm is included)
+  when: distro_supported|success
+  apt: pkg=nodejs state=latest update_cache=true
```
jmchilton: Seems great to me - thanks for updating node.

templates/supervisor.conf.j2
```diff
@@ -112,9 +112,9 @@ priority        = 200
...
+command         = {{ galaxy_venv_dir }}/bin/uwsgi --virtualenv {{ galaxy_venv_dir }} --yaml {{ galaxy_config_file }} --logdate --master --processes {{ galaxy_web_processes }} --threads {{ uwsgi_threads }} --logto {{ uwsgi_log }} --socket 127.0.0.1:{{ uwsgi_port }} --pythonpath lib --stats 127.0.0.1:9191 -b 16384
 {% else %}
-command         = {{ galaxy_venv_dir }}/bin/uwsgi --virtualenv {{ galaxy_venv_dir }} --ini-paste {{ galaxy_config_file }} --logdate --master --processes %(ENV_UWSGI_PROCESSES)s --threads %(ENV_UWSGI_THREADS)s --logto {{ uwsgi_log }} --socket 127.0.0.1:{{ uwsgi_port }} --pythonpath lib --stats 127.0.0.1:9191 -b 16384
+command         = {{ galaxy_venv_dir }}/bin/uwsgi --virtualenv {{ galaxy_venv_dir }} --yaml {{ galaxy_config_file }} --logdate --master --processes %(ENV_UWSGI_PROCESSES)s --threads %(ENV_UWSGI_THREADS)s --logto {{ uwsgi_log }} --socket 127.0.0.1:{{ uwsgi_port }} --pythonpath lib --stats 127.0.0.1:9191 -b 16384
```
mvdbeek: So far this is the only `.ini` incompatible change, we could introduce sth like `GALAXY_CONFIG_TYPE` and default that to `yaml` with the possibility to switch back to `ini-paste`

bgruening:
@drosofff @afgane does this look good to you?
It seems to work well for 18.01, as we are using it in the Docker builds.

tasks/nginx.yml
```diff
@@ -5,19 +5,19 @@
...
-    owner=root group=root mode=0600
+    owner={{ item.owner }} group={{ item.owner }} mode=0600
   with_items:
-    - { src: 'htpasswd.j2',                     dest: '/etc/nginx/htpasswd' }
-    - { src: 'nginx_galaxy_web.conf.j2',        dest: '{{ nginx_conf_directory }}/galaxy_web.conf' }
-    - { src: 'nginx_uwsgi.conf.j2',             dest: '{{ nginx_conf_directory }}/uwsgi.conf' }
-    - { src: 'nginx_reports_noauth.conf.j2',    dest: '{{ nginx_conf_directory }}/reports_auth.conf' }
-    - { src: 'nginx_reports_auth.conf.j2',      dest: '{{ nginx_conf_directory }}/reports_auth.conf.source' }
-    - { src: 'nginx.conf.j2',                   dest: '{{ nginx_conf_path }}' }
+    - { src: 'htpasswd.j2',                     dest: '/etc/nginx/htpasswd',                                 owner: 'galaxy' }
```
mvdbeek: What's the motivation for this ? I don't think things in /etc should be writable by the galaxy user

templates/supervisor.conf.j2
```diff
@@ -112,9 +112,9 @@ priority        = 200
 [program:galaxy_web]
 {% if galaxy_uwsgi|bool %}
 {% if galaxy_uwsgi_static_conf|bool %}
-command         = {{ galaxy_venv_dir }}/bin/uwsgi --virtualenv {{ galaxy_venv_dir }} --ini-paste {{ galaxy_config_file }} --logdate --master --processes {{ galaxy_web_processes }} --threads {{ uwsgi_threads }} --logto {{ uwsgi_log }} --socket 127.0.0.1:{{ uwsgi_port }} --pythonpath lib --stats 127.0.0.1:9191 -b 16384
+command         = {{ galaxy_venv_dir }}/bin/uwsgi --virtualenv {{ galaxy_venv_dir }} --yaml {{ galaxy_config_file }} --logdate --thunder-lock --master --processes {{ galaxy_web_processes }} --threads {{ uwsgi_threads }} --logto {{ uwsgi_log }} --socket 127.0.0.1:{{ uwsgi_port }} --pythonpath lib --stats 127.0.0.1:9191 -b 16384
```
mvdbeek: how about a variable `galaxy_config_style` for this ?

bgruening: You mean for yaml/ini? I wonder if we really want to add this complexity (there are a few more places where this has impact) or if we just go ahead.

mvdbeek: I think it is, it is very little addedcomplexity with the advantage of simple galaxy release switches, and I don't think in ansible-galaxy-extras there would be any other side-effect if you're not using those startup scripts that we don't use anyway.

bgruening:
Ok, green again! This time we are tracking all master branches again for testing and we can not spin up `docker swarm` htcondor clusters and scale them as we like.

jmchilton:
Woohoo - great work all, what an impressive collaboration!

templates/supervisor.conf.j2
drosofff: @bgruening Indeed `--yaml` not surprisingly breaks old style galaxy deployment. I'll definitely test the @mvdbeek suggestion to get GalaxyKickstart running back.
By the way, is there any ini2yaml converter that could wrap the old ini config files ?

mvdbeek: There's `make config-convert-dry-run` and `make config-convert` included with galaxy (which calls `python lib/galaxy/webapps/config_manage.py config-convert`)

PR #193, opened 2018-03-16
Fix broken _x_accel_redirect and proxy_prefix

abretaud:
Doing some tests with galaxy-stable 17.09, I had problems with proxy_prefix and _x_accel_redirect.
It turns out it is a consequence of https://github.com/galaxyproject/ansible-galaxy-extras/pull/185, so here's a fix for it = remove the proxy_prefix from _x_accel_redirect in galaxy config too

I don't know yet if 18.01 was impacted too. I guess yes.

@drosofff can you confirm it's ok for you too?


mvdbeek:
(it's a bit late now, but we're not at all using this file)

drosofff:
@abretaud it's ok for GalaxyKickStart. Thanks.

bgruening:
Thanks @abretaud!

PR #194, opened 2018-03-17
Support PROXY_PREFIX in 18.01

abretaud:
With this PR PROXY_PREFIX should now work with the new galaxy.yml
Cheers

### DIFF
--- a/tasks/nginx.yml
+++ b/tasks/nginx.yml
@@ -5,19 +5,29 @@
     state: directory
     path: "{{ nginx_upload_store_path }}"
     owner: "{{ galaxy_user_name }}"
-    
+
 - name: Place configuration files
   template: >
     src={{ item.src }}
     dest={{ item.dest }}
     owner=root group=root mode=0600
   with_items:
-    - { src: 'htpasswd.j2',                     dest: '/etc/nginx/htpasswd' }
     - { src: 'nginx_galaxy_web.conf.j2',        dest: '{{ nginx_conf_directory }}/galaxy_web.conf' }
     - { src: 'nginx_uwsgi.conf.j2',             dest: '{{ nginx_conf_directory }}/uwsgi.conf' }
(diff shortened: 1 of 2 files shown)


tasks/nginx.yml
```diff
@@ -5,14 +5,13 @@
...
     owner: "{{ galaxy_user_name }}"
-    
+
 - name: Place configuration files
   template: >
     src={{ item.src }}
     dest={{ item.dest }}
     owner=root group=root mode=0600
   with_items:
-    - { src: 'htpasswd.j2',                     dest: '/etc/nginx/htpasswd' }
```
bgruening: Does it work if you add  `owner: 'galaxy'` to this dict and use it above in the template?

templates/startup.sh.j2
```diff
@@ -75,29 +90,18 @@ if [ "x$PROXY_PREFIX" != "x" ]
         then
             export GALAXY_CONFIG_NGINX_UPLOAD_PATH="${PROXY_PREFIX}${GALAXY_CONFIG_NGINX_UPLOAD_PATH}"
     fi
-
-    if [ "$GALAXY_CONFIG_NGINX_X_ACCEL_REDIRECT_BASE" == "/_x_accel_redirect" ]
```
bgruening: Do we still need to maintain this for the ini case?

bgruening: ping @abretaud

tasks/nginx.yml
abretaud: It would look better yep, shall I make a PR for this?
Sorry for the delay, I've been sick the last days (seems like a galaxy meeting malediction for me!)

templates/startup.sh.j2
abretaud: No need to keep it, it's working the same for ini and yml

tasks/nginx.yml
bgruening: Yes please, not need to feel stressed, all good, get well soon.

PR #195, opened 2018-03-20
Adds thunder-lock to avoid race conditions between multiple uwsgi threads

pcm32:
This PR addresses the issue described in https://github.com/bgruening/docker-galaxy-stable/issues/423. I would love to hear if you have been avoiding the `--thunder-lock` option for some reason, as in my local and OpenStack tests this looks good and seems to avoid the race condition previously described.

@natefoo any thoughts about usage of `--thuder-lock`?

Thanks!

### DIFF
--- a/templates/supervisor.conf.j2
+++ b/templates/supervisor.conf.j2
@@ -112,9 +112,9 @@ priority        = 200
 [program:galaxy_web]
 {% if galaxy_uwsgi|bool %}
 {% if galaxy_uwsgi_static_conf|bool %}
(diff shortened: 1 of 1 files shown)


natefoo:
We have it [set on usegalaxy.org](https://github.com/galaxyproject/usegalaxy-playbook/commit/01f5d5af72269378d8f333f35d6d6d0c31c15bfd). I forget the exact conditions that brought it about, but it [seems like](https://uwsgi-docs.readthedocs.io/en/latest/articles/SerializingAccept.html) it should be set any time you are running with >1 process and >1 thread.

bgruening:
@natefoo is there a reason why we don't use this by default?

jmchilton:
I suspect it should be on by default but @natefoo isn't sold. I did create a PR to at least mention it and when to enable it in the default config https://github.com/galaxyproject/galaxy/pull/5743 though.

natefoo:
Just had a conversation with @jmchilton about this in the office. We're going to enable it by default if `processes` is set. For the default out-of-the-box config w/ no `galaxy.yml` it will not be set.

natefoo:
@jmchilton ha. Yeah, ok, I had that a bit wrong - it's not enabled by default but it's documented that it should be enabled if you set `processes > 1`. The reason it's not enabled by default is that in order to minimize cross-platform and site-specific incompatibilities for dev clones, it's best to only set the options that are absolutely necessary.

PR #196, opened 2018-03-21
PROXY_PREFIX + 18.01, 2nd

abretaud:
So here's the little change we talked about in https://github.com/galaxyproject/ansible-galaxy-extras/pull/194


bgruening:
Thanks :)

PR #199, opened 2018-05-01
Update nginx.yml

rhpvorderman:
I believe this issue was already solved in `master` branch.  Hardcoding the name 'galaxy' causes issues if galaxy_user_name is different.

### DIFF
--- a/tasks/nginx.yml
+++ b/tasks/nginx.yml
@@ -12,7 +12,7 @@
     dest={{ item.dest }}
     owner={{ item.owner }} group={{ item.owner }} mode=0600
   with_items:
-    - { src: 'htpasswd.j2',                     dest: '/etc/nginx/htpasswd',                                 owner: galaxy_user_name }
+    - { src: 'htpasswd.j2',                     dest: '/etc/nginx/htpasswd',                                 owner: '{{ galaxy_user_name }}' }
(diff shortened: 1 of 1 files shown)


tasks/nginx.yml
```diff
@@ -12,7 +12,7 @@
     dest={{ item.dest }}
     owner={{ item.owner }} group={{ item.owner }} mode=0600
   with_items:
-    - { src: 'htpasswd.j2',                     dest: '/etc/nginx/htpasswd',                                 owner: 'galaxy' }
```
bgruening: do you mean `{{ galaxy_user_name }}`

mvdbeek:
If this is solved in the master branch we should cherry-pick the commit(s) that solved the issue

tasks/nginx.yml
rhpvorderman: Yes

rhpvorderman:
Well I checked, and it is not solved in master branch.
In master branch files all these files should be owned by root.
In this branch /etc/nginx/htpasswd is owned by galaxy. This change was introduced in [this commit](https://github.com/galaxyproject/ansible-galaxy-extras/commit/37de7e8dea2334ddb031db9f4716547c67b7dd25).

tasks/nginx.yml
rhpvorderman: Solved this issue

bgruening:
cool thanks!

Issue #198, opened 2018-04-28
npm proxy dependencies installation failing


pcm32:
Hi all!

Currently installation of npm packages for proxy is failing like this:

```
TASK [galaxyprojectdotorg.galaxyextras : Install proxy dependencies.] **********
[...]

PR #200, opened 2018-05-01
Trust insecure npm

pcm32:
This allows allows building images again after issues described in #198, produced due to some issue in npm repo's certificate. I reckon this should be temporal and we might want to revert this later.


bgruening:
Thanks @pcm32!

PR #205, opened 2018-05-24
Update for galaxyproject/galaxy#5094.

jmchilton:
I think packed has been deprecated as long as this project has existed.


dannon:
Yep, this should be all it needs.

PR #207, opened 2018-05-31
Propose change to keep-alive timeout nginx.

rhpvorderman:
When managing dependencies in the admin panel, conda runs in the background. This might take longer than 65 seconds, at which point a gateway timeout follows.
Lets increase the keep-alive timeout. I set 600 for now, but another default should be more sane.

### DIFF
--- a/templates/nginx.conf.j2
+++ b/templates/nginx.conf.j2
@@ -16,7 +16,7 @@ http {
 
     sendfile        on;
 
-    keepalive_timeout  65;
+    keepalive_timeout  600;
     types_hash_max_size 2048;
 
     gzip  on;


mvdbeek:
I think @natefoo was arguing against this some time ago, but I may be wrong. Obviously the right thing would be to do the installation in a thread and return immediately. Which also applies to repository installations in general, which can also timeout.

rhpvorderman:
I agree with @mvdbeek. This is quite an ugly fix for a problem that should be solved upstream. I will create an issue on the galaxy repo. Unless an issue is already there.

rhpvorderman:
https://github.com/galaxyproject/galaxy/issues/6254

PR #210, opened 2018-06-22
Support uWSGI mule job handlers

natefoo:


mvdbeek:
https://github.com/galaxyproject/ansible-galaxy-extras/pull/211 should fix the test error.

PR #213, opened 2018-07-06
Failsafe option when galaxy config file is not in the expected place.

pcm32:
An old commit that I failed to PR, which sorts out issue https://github.com/bgruening/docker-galaxy-stable/issues/433 at docker-galaxy-stable.

### DIFF
--- a/templates/startup.sh.j2
+++ b/templates/startup.sh.j2
@@ -34,6 +34,14 @@ echo "127.0.0.1      galaxy" >> /etc/hosts
 # Set number of Galaxy handlers via GALAXY_HANDLER_NUMPROCS or default to 2
 ansible localhost -m ini_file -a "dest=/etc/supervisor/conf.d/galaxy.conf section=program:handler option=numprocs value=${GALAXY_HANDLER_NUMPROCS:-2}" &> /dev/null
 
+# If the Galaxy config file is not in the expected place, copy from the sample
+# and hope for the best (that the admin has done all the setup through env vars.)
+if [ ! -f $GALAXY_CONFIG_FILE ]
+  then
+  # this should succesfully copy either .yml or .ini sample file to the expected location
+  cp /export/config/galaxy${GALAXY_CONFIG_FILE: -4}.sample $GALAXY_CONFIG_FILE
+fi
+
 # Configure proxy prefix filtering
 if [ "x$PROXY_PREFIX" != "x" ]
     then


pcm32:
Hi there! I need this added to the main branch for 18.05 as well... I'm starting to use 18.05 on k8s and still seeing this issue:

```
IOError: [Errno 2] No such file or directory: '/etc/galaxy/galaxy.yml'
```

Thanks!

pcm32:
I'm a bit troubled by this. I recently started testing 18.05 on k8s, but I keep getting some HTTP errors (welcome page doesn't show up, file uploads fail with `Warning: Not Found (404)`), and before that I was getting some red errors at the very beginning (they went away with me setting galaxy.filter-with=/galaxy, although from documentation I'm not sure that this is even correct to do). 

On one side, the dagobah training slides it state that a `galaxy.yaml` file is not needed... yet without it (and the commit on this PR), galaxy 18.05 doesn't start on the compose setup (at least on k8s, maybe I'm doing something wrong). On the other hand, using the default yaml file, as proposed here, seems to be imposing some defaults for UWSGI that I'm not sure are correct, and that I'm starting to think might be related to my http errors... 

Any comments @bgruening ?

bgruening:
I'm not sure what is going on there. The yml should not be needed as Galaxy will fallback to the sample file. Also no configuration is done with the yaml file, everything should happen through supervisor and ENVs. I'm super confused here :(

And I have trouble to test this, as it seems to work for me with compose :(

pcm32:
Maybe we are building the containers with slightly different flags that in one case make certain scripts try to read the yaml file?

pcm32:
My http part setup is working, and looking at the default yaml setup there is no much there, so I presume that this is not affecting that part.

pcm32:
I see this again on 18.09 with a completely clean installation.... complete error is:

```
Waiting for database
Traceback (most recent call last):
  File "/usr/local/bin/check_database.py", line 18, in <module>
    db_url = get_config(sys.argv)['db_url']
  File "/export/galaxy-central/lib/galaxy/model/orm/scripts.py", line 126, in get_config
    properties = load_app_properties(config_file=config_file, config_prefix=config_override, config_section=config_section)
  File "/export/galaxy-central/lib/galaxy/util/properties.py", line 91, in load_app_properties
    with open(config_file, "r") as f:
IOError: [Errno 2] No such file or directory: '/etc/galaxy/galaxy.yml'
```

pcm32:
This has been merged by some other PR, as changes are there already:

https://github.com/galaxyproject/ansible-galaxy-extras/commit/934af939d4baf208fbed8edeafee40a06a47fe78#diff-205c98235541487412061d12953dc4c0

will close then this PR.

PR #214, opened 2018-07-13
Update main.yml

lecorguille:
My 2 cents

I was walking around and spot this:
According to this line: `galaxy_minimum_version: "17.01"` we shouldn't set the galaxy config file at `yaml` by default


abretaud:
Closing as galaxy_minimum_version is now 19.01
https://github.com/galaxyproject/ansible-galaxy-extras/pull/229/commits/fa1a3f8d00158d2a8486d72c85f33390f35d1e3b

Issue #216, opened 2018-09-22
Fetch DRMAA egg for Galaxy is now failing


drosofff:
see https://travis-ci.org/galaxyproject/ansible-galaxy-extras/jobs/395730003#L3753

And in command line I could reproduce the issue with the following output:

```
gks:~/galaxy$ .venv/bin/python scripts/fetch_eggs.py -e drmaa -c galaxy/config/galaxy.ini

Eggs in this release of Galaxy have been replaced by Python's newer packaging
format, wheels. Please use scripts/common_startup.sh to set up your
environment:

cd /home/galaxy/galaxy && ./scripts/common_startup.sh

This will create a Python virtualenv and install Galaxy's dependencies into it.

If you start Galaxy using means other than run.sh (as you probably do if you
are seeing this message), be sure to activate the virtualenv before starting,
using:

. /home/galaxy/galaxy/.venv/bin/activate

If you already run Galaxy in its own virtualenv, you can reuse your existing
virtualenv with:

cd /home/galaxy/galaxy && ./scripts/common_startup.sh --skip-venv

```

PR #217, opened 2018-09-23
fix issue with fetching eggs/wheels

drosofff:
as reported in https://github.com/galaxyproject/ansible-galaxy-extras/issues/216

### DIFF
--- a/tasks/slurm.yml
+++ b/tasks/slurm.yml
@@ -52,7 +52,7 @@
 
 - name: Check if Galaxy uses eggs or wheels
   stat:
-    path: "{{ galaxy_server_dir }}/lib/galaxy/dependencies/requirements.txt"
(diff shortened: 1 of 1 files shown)


tasks/slurm.yml
```diff
@@ -52,7 +52,7 @@
 
 - name: Check if Galaxy uses eggs or wheels
   stat:
-    path: "{{ galaxy_server_dir }}/lib/galaxy/dependencies/requirements.txt"
+    path: "{{ galaxy_server_dir }}/lib/galaxy/dependencies/pinned-requirements.txt"
```
mvdbeek: At this point it's so long ago that Galaxy used eggs, we're probably better off dropping egg support. I don't think it'd even work to run Galaxy 15.X with the other assumptions that this role makes.

PR #218, opened 2018-09-23
Drop egg support in tasks/slurm.yml

drosofff:
@mvdbeek (in response to https://github.com/galaxyproject/ansible-galaxy-extras/pull/217):

something like that ?

### DIFF
--- a/defaults/main.yml
+++ b/defaults/main.yml
@@ -98,7 +98,6 @@ galaxy_source_shellrc: false
 galaxy_user_shellrc: "{{ galaxy_home_dir }}/.bashrc"
 galaxy_log_dir: "{{ galaxy_home_dir }}/"
 galaxy_venv_dir: "{{ galaxy_home_dir }}/venv"
-galxy_egg_cache: "{{ galaxy_home_dir }}/.python-eggs"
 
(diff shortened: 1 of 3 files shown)


mvdbeek:
Yes, can you also drop the `galxy_egg_cache` variable in defaults/main.yml and the fetch_eggs step in tasks/pbs.yml ?

drosofff:
Should we address the issue of the test test failing at some point ?

mvdbeek:
Yes, I did look into this a bit and the proper fix would be to upgrade python in the container, but #219 should have fixed the tests

drosofff:
@mvdbeek agree, this is puzzling !!
maybe drop the the `dest=/dev/null` in `url="http://deb.nodesource.com/node/dists/{{ ansible_distribution_release }}/Release" dest=/dev/null` ? The @bgruening commit https://github.com/galaxyproject/ansible-galaxy-extras/commit/9e6bd6daa77f57dadc09ebb9bf00956248361d73 was also doing it.

drosofff:
and now the test of this PR passes. don't get it...

mvdbeek:
The default travis test merges the PR into master and then runs the tests on the result of the merge.

PR #220, opened 2018-12-12
[18.09] fix cvmfs install

abretaud:
I'm trying to fix errors in the galaxy docker compose images (a PR should come soon).
I found this problem with apt install of cvmfs, I *think* the key needs to be installed before adding the apt repo

### DIFF
--- a/tasks/cvmfs_client.yml
+++ b/tasks/cvmfs_client.yml
@@ -4,7 +4,9 @@
 
 # Install autofs
 - name: Install autofs system package
-  apt: pkg=['autofs', 'uuid-runtime'] state={{ galaxy_extras_apt_package_state }}
+  apt:
+    name: ['autofs', 'uuid-runtime']
+    state: "{{ galaxy_extras_apt_package_state }}"
   when: galaxy_extras_install_packages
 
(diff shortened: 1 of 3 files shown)


drosofff:
Hi Anthony,

I profit from the occasion, to propose that the ansible version is "pinned" and the same for all galaxyproject-ansible repositories. 
I am not saying that it is your issue, but I suspect pip install ansible without == in the .travis.yml is a possible option. 
On the same line I am currently investigating why the last PRs of ansible-galaxy-os are not compatible with GalaxyKickStart and I suspect similarly that this is a problem of ansible syntaxe.

Cheers

drosofff:
@bgruening 
> Good project for next year or maybe Xmas :)

Definitely for the next year, IntergalacticAnsible Utilities Commission IAUC ;-) I'd love !
For Xmas, I have unfortunately already several good projects waiting in the queue :-/

bgruening:
I will get this in, we can fix the failure in the branch or when we merge the branch back to master.

Thanks @abretaud!

bgruening:
@drosofff that was more a note to myself :)

PR #221, opened 2018-12-13
[18.09] small fixes

abretaud:
2 fixes for errors found in https://github.com/bgruening/docker-galaxy-stable/pull/464
- make sure cron is installed when it is managed by supervisor
- make sure the nginx upload module is not enabled when asking for https support. It is [disabled in http mode](https://github.com/bgruening/docker-galaxy-stable/blob/edbef97c5fa8f4d9d0f8e54db72b85bccba6d51e/compose/galaxy-web/Dockerfile#L61), but it was not specified here and it's [enabled by default](https://github.com/galaxyproject/ansible-galaxy-extras/blob/18.09/defaults/main.yml#L5).

### DIFF
--- a/defaults/main.yml
+++ b/defaults/main.yml
@@ -2,7 +2,7 @@
 # https://github.com/bgruening/docker-galaxy-stable.
 
 galaxy_extras_config_nginx: true
-galaxy_extras_config_nginx_upload: true
+galaxy_extras_config_nginx_upload: false
 galaxy_extras_config_postgres: true
 galaxy_extras_config_proftpd: true
 galaxy_extras_config_slurm: true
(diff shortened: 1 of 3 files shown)


abretaud:
I think the planemo-machine test is failing for the same reason: it tries to enable the upload_store because it is enabled by default, and it fails starting nginx as the module is not available.
Maybe I could make it disabled by default in https://github.com/galaxyproject/ansible-galaxy-extras/blob/18.09/defaults/main.yml#L5?
I could also disable it in [planemo-machine](https://github.com/galaxyproject/planemo-machine/blob/master/group_vars/all) I guess

bgruening:
I think disable it by default is ok.

bgruening:
Muha, thanks so much!

PR #230, opened 2019-03-26
adding Ubuntu 18.04 compatibility

luke-c-sargent:
This PR addresses a few issues with using this component of the GalaxyKickstart stack:
1. using `ppa:natefoo/slurm-drmaa` to get slurm components
2. slurm-wlm in lieu of slurm-llnl when using Ubuntu 18.04
3. fix in ` templates/configure_slurm.py.j2` to deal with some cruft left in the 'split' mechanism of status parsing

this has been tested on Ubuntu 18.04; i was able to install some tools from the shed and run [Galaxy 101](https://galaxyproject.org/tutorials/g101/) to completion. When using it as part of GalaxyKickstart, I successfully installed to Ubuntu 16.04 and 18.04, though it required some modifications to galaxy-os which i plan to issue a PR to should this approach be deemed acceptable.

critiques welcome!

### DIFF
--- a/tasks/galaxy_extra_dependencies.yml
+++ b/tasks/galaxy_extra_dependencies.yml
@@ -4,7 +4,7 @@
     name: "{{item}}"
     virtualenv: "{{ galaxy_venv_dir }}"
     virtualenv_command: "{{ pip_virtualenv_command | default( 'virtualenv' ) }}"
-    extra_args: --index-url https://wheels.galaxyproject.org/simple
+    extra_args: "--index-url https://wheels.galaxyproject.org/simple  --extra-index-url https://pypi.python.org/simple"
   become: True
   become_user: "{{ galaxy_user_name }}"
   with_items:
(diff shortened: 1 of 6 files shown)


drosofff:
Hi @luke-c-sargent. What are the modification planned in ansible-galaxy-os ?

drosofff:
I see @luke-c-sargent !
```
No package matching 'slurm-drmaa-dev'
```
It work effortlessly in ubuntu 16.04, so at least this part of deployment in galaxykickstart is preserved ;-)
For the rest, I guess it will be ok with the change in galaxy-os

drosofff (approved):

luke-c-sargent:
@drosofff 
the mentioned galaxy-os PR:
https://github.com/galaxyproject/ansible-galaxy-os/pull/29

it's the same repo addition, since that role also installs slurm-drmaa-dev; i assume the redundancy is for flexibility, in case you want to use either one role or the other, but not both (irrelevant for GKS).

Side note: 14.04 installations no longer work; this is probably for the best since its end of life is next month, but many of the ansible-galaxy-* roles mention it as the preferred Ubuntu version, which should probably be changed.

abretaud:
Could you rebase/fix the conflicts? There's been some change on the master including things overlapping with this PR + the tests are now run with ubuntu 18.04

PR #233, opened 2019-04-09
Missing setting to allow REMOTE_USER and GX_SECRET

innovate-invent:


dannon:
The build failure is unrelated to this PR, I think?

bgruening:
Yes, they are. @innovate-invent do you want to have this in 19.01 or in the next 19.05? If the latter please apply this to the 19.05 branch which I started already. If you want it in 19.01 then please use the master branch.

Thanks!

innovate-invent:
I expect so. The build log is too long.

innovate-invent:
It would be good to backport this, changed to master.

innovate-invent:
This change does not completely allow REMOTE_USER to be used.
[this line](https://github.com/galaxyproject/ansible-galaxy-extras/blob/36c5b7b7dc49c14a78a91d3f063d927262d1c251/templates/nginx_uwsgi.conf.j2#L11) interferes and must be removed to allow access. I am not sure is it is needed by something.

abretaud:
Ah sorry @innovate-invent I hadn't seen this other PR and your last comment. I merged it in master in 3c6fe3585aa3b82d22f74a1c3eb59afc1a2a49c0

I don't really understand why https://github.com/galaxyproject/ansible-galaxy-extras/blob/36c5b7b7dc49c14a78a91d3f063d927262d1c251/templates/nginx_uwsgi.conf.j2#L11 is conflicting with your change, do you have more details?

innovate-invent:
Sorry, that was a year ago and I can't remember. I think it may have had to do with the way nginx behaves and that line overrides forwarding all headers. I am not sure.

Issue #235, opened 2019-12-08
Installation of watchdog and drmaa are failing


drosofff:
Most probably as a consequence of https://twitter.com/natefoo/status/1202979010860474373
@natefoo does it make sense ?

abretaud:
It looks like this is fixed, reopen if not

PR #236, opened 2019-12-08
fix problems with broken wheels.galaxyproject.org

drosofff:
This PR is a turn around of the  issue #235

### DIFF
--- a/tasks/galaxy_root.yml
+++ b/tasks/galaxy_root.yml
@@ -11,9 +11,9 @@
 - name: "Install watchdog for galaxy"
   pip:
     name: "watchdog"
+    version: 0.9.0
     virtualenv: "{{ galaxy_venv_dir }}"
     virtualenv_command: "{{ pip_virtualenv_command | default( 'virtualenv' ) }}"
-    extra_args: --index-url https://wheels.galaxyproject.org/simple
   become: True
   become_user: "{{ galaxy_user_name }}"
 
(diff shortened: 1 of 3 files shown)


tasks/galaxy_root.yml
```diff
@@ -11,9 +11,9 @@
 - name: "Install watchdog for galaxy"
   pip:
     name: "watchdog"
+    version: 0.9.0
```
mvdbeek: I think you need to fix the pip extra args as in https://github.com/galaxyproject/ansible-galaxy/blob/master/tasks/dependencies.yml#L14 ... but I think with a recent ansible-galaxy you don't need these tasks anymore. I would skip this entirely, but at the minimum you can't pin the watchdog version here.

tasks/slurm.yml
```diff
@@ -63,7 +63,7 @@
 - name: Fetch DRMAA wheel for Galaxy
   pip:
     name: "drmaa"
-    extra_args: "--index-url https://wheels.galaxyproject.org/simple/"
+    version: 0.7.9
```
mvdbeek: Same here, we shouldn't pin any versions here, fixing the extra_args should be sufficient to get this to run.

bgruening:
We need to merge in the 19.05 branch. This one already fixes some things of 19.09 including wheels.

drosofff:
@bgruening Yes, it isn't even necessary to ever merge this branch if the issue is better addressed in 19.05. It was a kind of emergency patch for a galaxy training.

Issue #237, opened 2019-12-13
Fix role testing


drosofff:
Hi @bgruening 
I am trying to fix a little bit the tests of the extras role.
What I am trying to figure out currently is how the galaxy-postgresql ansible roles is downloaded during [ci_test_docker_galaxy.sh](https://github.com/galaxyproject/ansible-galaxy-extras/blob/master/ci_test_docker_galaxy.sh) execution, as requirement for the docker build of the galaxy-docker/test image.
In facts, I see that the docker build is failing when the playbook task "Stop and remove postgresql" is run
```
TASK [galaxyprojectdotorg.galaxyextras : Stop and remove postgresql.] **********
failed: [localhost] (item=postgresql) => {"changed": false, "item": "postgresql", "msg": "Could not find the requested service postgresql: host"}

RUNNING HANDLER [galaxyprojectdotorg.galaxyextras : restart nginx] *************
	to retry, use: --limit @/ansible/provision.retry

PLAY RECAP *********************************************************************
localhost                  : ok=43   changed=34   unreachable=0    failed=1
```
And I can also see that it is expected since this role is empty in the container, maybe (but not sure at all) because the github submodule galaxy-postgresql had not been explicitely included in the test code.

PR #238, opened 2019-12-21
Improve repo testing (introducing a galaxykickstart job)

drosofff:
Here is an attempt to improve the automatic tests for this repo.
- Since the `ci_test_docker_galaxy.sh` is failing since several months, I introduced a similar procedure which instead of being based on galaxy-docker-stable is based on [galaxykickstart](https://github.com/ARTbio/GalaxyKickStart.git).

- Before that, I had also worked a bit on the `ci_test_docker_galaxy.sh` issues and I kept some fixes in this PR (in particular the postgresql is now installed, see #237 ). There are still some fixes to be found but I did not had time to do it. To whom it may concern, I identified a strange issue with the apt-get install of condor package.

- The PR is doing a regular pip install of watchdog and slurm, as a turn around of #235. I saw that @bgruening is currently working on that and I will be happy to pick any commits that address the issue better than my quick and dirty patch.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,11 +1,14 @@
 ---
 language: python
-python: "2.7"
+python: 2.7
+dist: xenial
+sudo: required
 
 env:
   global:
     - TOX_ENV=py27
   matrix:
+    - SUITE=galaxykickstart
     - SUITE=docker-galaxy
     - SUITE=planemo-machine
     - SUITE=syntax
@@ -27,7 +30,7 @@ before_install:
 
 install:
   # Install Ansible.
-  - pip install ansible
+  - pip install ansible==2.7.4
 
   # Add ansible.cfg to pick up roles path.
   - printf '[defaults]\nroles_path = ../' > ansible.cfg
@@ -35,4 +38,5 @@ install:
 script:
   - if [ "$SUITE" == "syntax" ]; then ansible-playbook -i "localhost," tests/syntax.yml --syntax-check; fi
   - if [ "$SUITE" == "docker-galaxy" ]; then bash ci_test_docker_galaxy.sh; fi
+  - if [ "$SUITE" == "galaxykickstart" ]; then bash ci_test_galaxykickstart.sh; fi
   - if [ "$SUITE" == "planemo-machine" ]; then bash ci_test_planemo_machine.sh; fi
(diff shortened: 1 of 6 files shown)


.travis.yml
```diff
@@ -27,12 +30,13 @@ before_install:
 
 install:
   # Install Ansible.
-  - pip install ansible
+  - pip install ansible==2.7.4
```
mvdbeek: I think I'd consider ansible-galaxy-extras more as a library used by many projects. So if this role can run with the latest ansible I wouldn't pin ansible for the tests. This way we can have regular tests that will break when ansible change something in their code and we need to update.

ci_test_docker_galaxy.sh
```diff
@@ -2,8 +2,11 @@
 
 set -e
 
-mkdir $HOME/galaxy-docker
-wget -q -O - https://github.com/bgruening/docker-galaxy-stable/archive/master.tar.gz | tar xzf - --strip-components=1 -C $HOME/galaxy-docker
+# this failed because galaxy-postgresql is not recursively imported
```
mvdbeek: I'd leave this up to @bgruening or others to fix, I'm happy to merge without docker-galaxy tests passing, that seems to be work in progress.

.travis.yml
```diff
@@ -27,12 +30,13 @@ before_install:
...
 
   # Add ansible.cfg to pick up roles path.
   - printf '[defaults]\nroles_path = ../' > ansible.cfg
 
 script:
   - if [ "$SUITE" == "syntax" ]; then ansible-playbook -i "localhost," tests/syntax.yml --syntax-check; fi
   - if [ "$SUITE" == "docker-galaxy" ]; then bash ci_test_docker_galaxy.sh; fi
+  - if [ "$SUITE" == "galaxykickstart" ]; then bash ci_test_galaxykickstart.sh; fi
```
mvdbeek: very nice and clean, thanks!

.travis.yml
drosofff: OK, I agree

ci_test_docker_galaxy.sh
drosofff: @mvdbeek @bgruening should I revert this particular commit to leave the work for 19.05 ?

bgruening: Leave it in. I'm just surprised why this stopped working or ever worked...

jmchilton:
Awesome - thanks!

PR #239, opened 2020-01-09
Fix nginx conf for icon display in releases >=19.09

drosofff:
- I found these patches are enough to display correctly small icons in the `manage tool` panel.
Not sure it is the more elegant patch to adapt to the new path of these png files, but it does not hurt previous releases either

- The PR also fixes the ci_test_galaxykickstart.sh by using ansible 2.7.4 instead of ansible 2.9.2. A side info from this fix is that the ansible-galaxy-tools role should be fixed to accommodate ansible 2.9.2 - if desired (probably a problem of access rights differently managed by the two ansible versions).

### DIFF
--- a/ci_test_galaxykickstart.sh
+++ b/ci_test_galaxykickstart.sh
@@ -40,7 +40,7 @@ sudo su $GALAXY_TRAVIS_USER -c 'pip install --ignore-installed --user https://gi
 sudo -E su $GALAXY_TRAVIS_USER -c "export PATH=$GALAXY_HOME/.local/bin/:$PATH &&
 cd $GALAXY_HOME &&
 bioblend-galaxy-tests -v -k 'not test_collections_in_history_index and \
-              not test_create_list_in_history \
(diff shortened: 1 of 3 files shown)


dannon:
Is this the same problem as addressed in https://github.com/galaxyproject/galaxy-helm/pull/87/files?

drosofff:
@dannon I cannot say...

abretaud:
@drosofff could you rebase/resolve conflicts?
Tests are passing now on master, so I guess pining to an older ansible is no longer needed

drosofff:
Hi @abretaud I have a few hours to fix this. can you refresh my memory with several human readable sentences to summarize the situation: this pr is quite old and I confess that it is not in the center of my attention anymore...

abretaud:
Yes, looking closely at the diff, I don't think the changes are needed anymore:
- the tests are passing now
- there where some changes with the images path

So I'm closing this PR, feel free to open a new one if you find out that some changes are still needed

PR #245, opened 2020-10-25
Fix supervisor and proftpd tasks, and repo testing by gks

drosofff:
- Fix the supervisor.conf.j2 template, allowing nodejs to be found and started by supervisor
- Fix the proftpd.conf.j2, allowing proftpd to accommodate the current Galaxy password encoding
- connects with the new brand galaxykickstart (py3.7) for travis ci testing (this travis vm now turns green)
- my opinion in that docker testing could be suspended until it is fixed, but it is a question of taste (I prefer the green to the red) and not a call for that pr.

Feedbacks and improvements welcome before merging !

### DIFF
--- a/templates/supervisor.conf.j2
+++ b/templates/supervisor.conf.j2
@@ -202,7 +202,7 @@ command         = {{ galaxy_server_dir }}/lib/galaxy/web/proxy/js/lib/main.js --
 autostart       = {{ supervisor_ie_proxy_autostart }}
 autorestart     = unexpected
 user            = {{ galaxy_user_name }}
-environment     = PATH={{ galaxy_venv_dir }}/bin
+environment     = PATH={{ galaxy_venv_dir }}/bin:%(ENV_PATH)s
 startsecs       = 5
 redirect_stderr = true
 {% endif %}
--- a/templates/supervisor.conf.j2
+++ b/templates/supervisor.conf.j2
@@ -202,6 +202,7 @@ command         = {{ galaxy_server_dir }}/lib/galaxy/web/proxy/js/lib/main.js --
 autostart       = {{ supervisor_ie_proxy_autostart }}
 autorestart     = unexpected
 user            = {{ galaxy_user_name }}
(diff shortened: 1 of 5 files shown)


mvdbeek (approved):

templates/supervisor.conf.j2
```diff
@@ -202,6 +202,7 @@ command         = {{ galaxy_server_dir }}/lib/galaxy/web/proxy/js/lib/main.js --
 autostart       = {{ supervisor_ie_proxy_autostart }}
 autorestart     = unexpected
 user            = {{ galaxy_user_name }}
+environment     = PATH={{ galaxy_venv_dir }}/bin
```
mvdbeek, line 205: Should this maybe be 
```suggestion
environment     = PATH={{ galaxy_venv_dir }}/bin:%(ENV_PATH)s
```
This would also pull in the `PATH` environment variable of the process that started supervisor ?

drosofff: Sure ! I 'll do it.
I profit from the opportunity to catch you on an ansible topic:

I am scratching my head  on a change in ansible-galaxy which seems to occurred in July, although I am not 100% sure this is the repo to blame:

I can't have any more the posgresql database built with galaxykickstart, a bug that by the way is although visible in the test of this repository ansible-galaxy-extras (see [here](https://travis-ci.org/github/galaxyproject/ansible-galaxy-extras/jobs/738757721))

```
TASK [galaxyprojectdotorg.galaxy : Get current Galaxy DB version] **************
ok: [localhost]
TASK [galaxyprojectdotorg.galaxy : Get maximum Galaxy DB version] **************
ok: [localhost]
TASK [galaxyprojectdotorg.galaxy : Report current and max Galaxy database] *****
skipping: [localhost]
TASK [galaxyprojectdotorg.galaxy : Upgrade Galaxy DB] **************************
skipping: [localhost]                                                                    <--- this is not good !
```

It may well be also related to variable changes that I made in the GKS playbook, but I can't figure out which ones.

Would you have a clue to share ?

PR #248, opened 2021-04-30
Pythonize 3 and fix small bug for use in ubuntu 20.04

drosofff:
* Update supervisor.yml

* fix slurm install in 20.04. (install python3-pip instead of python-pip)

* Update proftpd.conf.j2 to fix the use of environment variables in the last versions of proftpd

### DIFF
--- a/tasks/supervisor.yml
+++ b/tasks/supervisor.yml
@@ -8,7 +8,18 @@
     - uwsgi
     - uwsgi-plugin-python
     - supervisor
-  when: galaxy_extras_install_packages
+  when: galaxy_extras_install_packages and ansible_distribution_version is version('20.04', '<')
+
+- name: Install web related system packages
+  apt: 
+    state: "{{ galaxy_extras_apt_package_state }}"
+    name: "{{ packages }}"
+  vars:
+    packages:
+    - uwsgi
+    - uwsgi-plugin-python3
+    - supervisor
+  when: galaxy_extras_install_packages and ansible_distribution_version is version('20.04', '==')
 
 - name: Install cron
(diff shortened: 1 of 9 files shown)


tasks/supervisor.yml
```diff
@@ -8,7 +8,18 @@
...
+  when: galaxy_extras_install_packages and ansible_distribution_version is version('20.04', '<')
+
+- name: Install web related system packages
+  apt: 
+    state: "{{ galaxy_extras_apt_package_state }}"
+    name: "{{ packages }}"
+  vars:
+    packages:
+    - uwsgi
+    - uwsgi-plugin-python3
```
mvdbeek: It's much better to install uwsgi from pip, the ubuntu versions are typically very out of date and will likely not work with modern Python 3 (>=3.7.1). This should be taken care of by the ansible-galaxy role, IIRC.

drosofff: All right @mvdbeek thanks for catching that pyuwsgi is indeed in the pinned_requirements of galaxy.
I have also added 2 github ci workflows to test the repo (based on galaxykickstart of course...), but I think an additional push of something is needed to trigger the github actions in it. Find something else to fix ;-)

drosofff: Indeed there is some ansible syntax to lint, this will trigger the actions ;-)

drosofff: forget, the github action, it is probably because it is only triggered on a push on the local repo. you'll have it if you merge... In the meantime, look at [this](https://github.com/ARTbio/ansible-galaxy-extras/actions)

